PR #4, opened 2022-01-12
Extrinsics decoding

tmcgroul:


substrate-metadata/src/extrinsic.ts
```diff
@@ -0,0 +1,24 @@
...
+
+const BIT_SIGNED = 0b10000000;
+
+export function decodeExtrinsic(rawExtrinsic: string, description: ChainDescription): Extrinsic {
+    let data = Buffer.from(rawExtrinsic.slice(2), 'hex')
+    let src = new Src(data)
+    let codec = new Codec(description.types)
+
+    src.compact()
+    let signed = (src.u8() & BIT_SIGNED) == BIT_SIGNED
```
eldargab, line 13: This byte looks a bit strange, because they could encode signatures as a regular optional field.
What is the meaning of other bits here? Were it is implemented in substrate?

tmcgroul: I have found a few resources which did the same:
https://docs.substrate.io/rustdocs/latest/src/sp_runtime/generic/unchecked_extrinsic.rs.html#210
https://github.com/polkascan/py-scale-codec/blob/964c33cb31dc1d089f6a3af23844b06b60ddd035/scalecodec/types.py#L2383
https://github.com/polkadot-js/api/blob/8ba4ed53776a7a71478e229547c7af3d4dc57d63/packages/types/src/extrinsic/Extrinsic.ts#L45 (haven't found actual byte reading)

tmcgroul:
@eldargab i've pushed a few little changes... Now it's ready to merge!

eldargab:
Merged in branch `sa-archive`

PR #20, opened 2022-03-14
feat: some config for saas

rightjelkin:
Some improvements for saas module.
It has become necessary to changing base API url because of development environment has been deployed.
Anyway, this feature would also need to be implemented (for the times, when we will have different saas providers)

Credential file was changed to simple yaml config with two fields (at this moment).

We can create a command for changing config later.

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -54,7 +54,8 @@
     "pgtools": "^0.3.2",
     "query-string": "^7.1.0",
     "simple-git": "^2.48.0",
-    "typeorm": "^0.2.41"
+    "typeorm": "^0.2.41",
+    "yaml": "~1.10.2"
   },
   "devDependencies": {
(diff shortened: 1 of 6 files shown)


rightjelkin:
> Why yaml for technical data? Why extra dependency?

Good question
I wanted to make this more user-friendly...may be it would be helpful for future development...

I don't mind rewriting it with JSON, if you think, that yaml is a kind of overengineering)

rightjelkin:
yaml npm package is zero dependency btw

eldargab:
Let's stick with JSON. I can't see what we can win here. 

Also all other configs in squid framework are JSON simply because the whole thing (validation, tolerant parser from Microsoft, etc) is much better maintained.

rightjelkin:
Ok, i will refactor it...let it be in same PR

PR #28, opened 2022-03-29
Add validator to SubstrateBlock

belopash:


dzhelezov:
Lgtm

substrate-processor/src/interfaces/substrate.ts
```diff
@@ -53,6 +53,11 @@ export interface SubstrateBlock {
      */
     lastRuntimeUpgrade: unknown
 
+    /**
+     * Hex string representing the block validator
+     */
+    validatorId?: string
```
eldargab, line 59: May be call it `validator`?

belopash: but yeah... just "validator" will be better

PR #32, opened 2022-04-13
Add in line documentation

RaekwonIII:


substrate-processor/src/interfaces/handlerContext.ts
```diff
@@ -4,12 +4,29 @@ import type {FindOneOptions} from "typeorm/find-options/FindOneOptions"
...
 
-
+/**
+ * Represents an interface to the database, can create, delete, fetch any {@link Entity} and run arbitrary queries.
+ * 
+ * Extends parent class by adding a {@link get} method.
+ */
 export interface Store extends EntityManager {
+    /**
+     * Uses the unterlying database interface to fetch an {@link Entity}, given the provided options, or its id.
```
dzhelezov, line 14: type: underlying

substrate-processor/src/batch.ts
```diff
@@ -18,13 +27,25 @@ export interface DataHandlers {
...
+ * @property handlers: a {@link DataHandlers} objects to collect the handlers that should be triggered for the batch.
+ */
 export interface Batch {
     range: Range
     handlers: DataHandlers
 }
 
-
+/**
+ * Forms an array of {@link Batch} for the given {@link Range} by mixing in the provided hooks
```
dzhelezov, line 42: This is a bit confusing as it's not clear what 'mixing' really means. Smth like `Extracts a Batch of handlers that should be executed within a given block Range`

substrate-processor/src/batch.ts
```diff
@@ -18,13 +27,25 @@ export interface DataHandlers {
     evmLogs: Record<EvmContractAddress, {filter?: EvmTopicSet[], handler: EvmLogHandler}[]>
 }
 
-
+/**
+ * Defines a batch of blocks to be processed
+ * 
+ * @property range: a {@link Range} defining start and end blocks for the batch
+ * @property handlers: a {@link DataHandlers} objects to collect the handlers that should be triggered for the batch.
```
dzhelezov, line 34: it should be stressed that each handler in a batch is admissible in the given range.

substrate-processor/src/batch.ts
```diff
@@ -125,7 +146,11 @@ export function createBatches(hooks: Hooks, blockRange?: Range): Batch[] {
     return batches
 }
 
-
+/**
+ * Given a list of {@link Batch}, it merges their Data Handlers by coalescing them based on their {@link Range}
```
dzhelezov, line 150: It's not clear what 'coalescing' means here.

substrate-processor/src/batch.ts
```diff
@@ -191,7 +220,14 @@ function mergeMaps<T>(a: Record<string, T>, b: Record<string, T>, mergeItems: (a
     return result
 }
 
-
+/**
+ * Given a list of {@link Range}, it counts the total number of blocks, adding each range until the chain height is
```
dzhelezov, line 224: It counts the total number of blocks spanned by the ranges, up to the chain height.

substrate-processor/src/chain.ts
```diff
@@ -34,7 +40,10 @@ interface BlockInfo {
     runtimeVersion: SpecVersion | SubstrateRuntimeVersion
 }
 
-
+/**
+ * Keeps a mapping of {@link Chain} linked to the blockchain's height, to encapsulate metadata in relation to the
```
dzhelezov, line 44: Sounds confusing to me. The Manager keeps track of spec versions and maps them to the segments of blocks for which the given spec version is constant. Such block segments (called Chains by Eldar here -- not really a good name imo) have well-defined metadata for events, extrinsics and storage items, so it is possible to decode them.

substrate-processor/src/ingest.ts
```diff
@@ -7,13 +7,23 @@ import {AbortHandle, Channel, wait} from "./util/async"
 import {hasProperties, unique} from "./util/misc"
 import {rangeEnd} from "./util/range"
 
-
+/**
+ * Defines block data as part of the ingestion inflow
```
dzhelezov, line 11: What does it mean "as part of the ingestion flow"? Probably you mean that this is the data retrieved for each block?

substrate-processor/src/interfaces/hooks.ts
```diff
@@ -3,36 +3,76 @@ import {EvmLogHandler, EvmTopicSet} from "./evm"
...
 export interface ExtrinsicHook {
     handler: ExtrinsicHandler
     event: QualifiedName
     extrinsic: QualifiedName
     range?: Range
 }
 
-
+/**
+ * Defines a {@link EvmLogHandler} function and the contract address that should restrain its execution.
```
dzhelezov, line 46: It should be noted when exactly EvmLogHandler is triggered. The following conditions must be satisfied:
- the evm log event was emitted by the given contract address
- the topic of the evm log event matches  `EvmTopicSet` as defined by the [evm event topic convention](https://docs.ethers.io/v5/concepts/events/#events--filters) (if present)
- the block height is in range (if present)

substrate-processor/src/interfaces/hooks.ts
```diff
@@ -3,36 +3,76 @@ import {EvmLogHandler, EvmTopicSet} from "./evm"
...
+ */
 export interface EvmLogHook {
     handler: EvmLogHandler
     contractAddress: string
     filter?: EvmTopicSet[]
     range?: Range
 }
 
-
+/**
```
dzhelezov, line 62: Here it should be noted that for each block the execution order is as follows:
- pre block hooks for this blocks
- event hooks are executed in the order the events are emitted within the block
- extrinsic hooks are executed in the order the triggerEvent ( `system.ExtrinsicSuccess` by default) are emitted within the block
- post block hooks are executed after all other hooks
- postHooks are executed after all the event and extrinsics in the block

substrate-processor/src/processor.ts
```diff
@@ -1,350 +1,640 @@
...
-import {Hooks} from "./interfaces/hooks"
-import {QualifiedName, SubstrateEvent} from "./interfaces/substrate"
-import {ProgressTracker} from "./progress-tracker"
-import {Prometheus} from "./prometheus"
-import {timeInterval} from "./util/misc"
-import {Range} from "./util/range"
-import {ServiceManager} from "./util/sm"
-
-
+import { ResilientRpcClient } from "@subsquid/rpc-client/lib/resilient";
```
dzhelezov, line 1: Too many edits due to the reformatting, let's not introduce it in this commit

substrate-processor/src/processor.ts
```diff
@@ -1,350 +1,640 @@
...
-            ...options
-        })
+    this.hooks.post.push({ handler, ...options });
+  }
+
+  /**
+   * Adds an {@link EventHandler} function that should be executed when an Event with the specified name is 
+   * encountered. Can accept an additional {@link EventHandlerOptions} parameter.
+   * 
+   * @param eventName a {@link QualifiedName} (a string), representing the name of the Event of interest. When such
```
dzhelezov, line 283: It should be specified what a QualifiedName is, that is, `${module}.{name}`. Should it be camel-cased?

substrate-processor/src/processor.ts
```diff
@@ -1,350 +1,640 @@
...
-        this.hooks.event.push({
-            event: eventName,
-            handler,
-            ...options
-        })
+    this.hooks.post.push({ handler, ...options });
+  }
+
+  /**
+   * Adds an {@link EventHandler} function that should be executed when an Event with the specified name is
```
dzhelezov, line 280: should be execution -> is executed

substrate-processor/src/processor.ts
```diff
@@ -1,350 +1,640 @@
...
+      new ChainManager(client, this.typesBundle),
+      db,
+      prometheus,
+      progress
+    );
+  }
+
+  /**
+   * Performs the actual processing the blockchain. Uses the {@link Ingest} class to obtain blocks in batches, then 
+   * triggers the configured pre/post block handlers, as well as any Event and Extrinsic handlers that should be
```
dzhelezov, line 501: ...in the canonical order defined by the ordering of the events in the block.

eldargab (changes requested):
Hey, guys, a couple of general notes!

1. First of all, there is no rush to comment on internal APIs which are not supposed to be used by the end-user. Most of the PR talks about internal stuff which is already out of date.

2. Tons of meaningless comments.

3. Don’t forget that bad documentation is much worse than not having documentation at all!

substrate-processor/src/batch.ts
```diff
@@ -6,7 +6,16 @@ import {QualifiedName} from "./interfaces/substrate"
 import {Heap} from "./util/heap"
 import {Range, rangeDifference, rangeIntersection} from "./util/range"
 
-
+/**
+ * Collects handlers for the various trigger types (pre/post block, event, extrinsic, EVM log).
+ * Keeps a list of handlers of the type corresponding to the trigger
+ * 
+ * @property pre: array of {@link BlockHandler}
```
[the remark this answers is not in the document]
RaekwonIII: The way you are proposing does not show up in an IDE when hovering over the class. It does not allow people to discover internal properties.

image

With `@properties`, they are visible from the class or interface. Since self-discovery is one of the main issues of our users...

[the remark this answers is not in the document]
eldargab: One is not supposed to see all class properties just by hovering over class or interface. Usually this is just too much.
Developers typically explore stuff via "jump to definition".

There is also another way:

image

image

substrate-processor/src/chain.ts
RaekwonIII: Definitely better wording.

substrate-processor/src/ingest.ts
RaekwonIII: Pretty much, yeah. I put myself in a corner by starting with "block data", instead of "data that..."

substrate-processor/src/batch.ts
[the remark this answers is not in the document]
dzhelezov: The main issue we want to address is to show the developers all possible options when developing mappings. It's virtually impossible to figure out that there options like 'setBatchSize' or 'triggerEvents'. IDE should help with that, so whenever you type 'processor...' all possible setters are visible with a short description. 

Same for handlers and context definitions -- the main goal is to make sure the developers are aware what is possible with Subsquid. The details can be then figured out via external docs or through the code (depending on how vigilant they are)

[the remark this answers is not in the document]
eldargab: image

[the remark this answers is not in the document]
RaekwonIII: > Developers typically explore stuff via "jump to definition".

I know I do, that's what I did to understand the project and write the docs, but judging on the question we get on a daily basis, not many others do. Unfortunately. 😢 

I understand what you are saying Eldar, the thing is that having Interfaces as function arguments "hides" what's inside it.

image

If we want people to self serve and do the documentation as you suggest (not that I am against it, in principle you are absolutely correct), they need to know what Range is, what it looks like. Ultimately, what they need to put as an argument to the function. (It's the same thing for `setDataSource`, by the way).

But I looked at the comment of the function itself, while writing this comment:

```
    /**
     * Sets a range of blocks to restrict processing of a blockchain. (optional)
     * When used, the processor will start extracting information from the specified `from` block in the {@link Range} 
     * object and (optionally) stop at the `to` indicated block.
     * 
     * If not set, the processor will default to start from block number 0 and never stop.
     * 
     * @param range an object with `from` and (optionally) `to` fields, defining the block numbers to which the chain 
     * processing should be restricted. @see Range
     */
    setBlockRange(range: Range): void {
        this.assertNotRunning()
        this.blockRange = range
    }
```

And it changed my mind a little bit. I tried to do it in one go, because I wanted to get other things done and probably by the end of it I was not lucid.

I am going to do another pass and try to make sure that functions are accessible and easy to use.

About what Dimitry said: my suggestion is that if methods like `setDataSource` are mandatory, they should not be "discovered" by doing `processor.` and waiting for the IDE to suggest it. Usually, mandatory settings are part of the constructor, or there are patterns like [Factory Method](https://en.wikipedia.org/wiki/Factory_method_pattern). You could even have what in Python are "class methods" that call the constructor to initialize parameters, like:

```typescript

export class SubstrateProcessor {

  getProcessorFromDataSource(name: string, src: DataSource): SubstrateProcessor {
    const processor = new SubstrateProcessor(name);
    processor.setDataSource(src);
    return processor;
  } 
}
```

Silly quick example, but I hope you get the idea.

PR #36, opened 2022-05-02
setup openreader schema for new archives

tmcgroul:


openreader/src/queryBuilder.ts
```diff
@@ -465,9 +485,15 @@ export class QueryBuilder {
...
                         break
                     }
-                    case 'list-lookup':
-                        rec[req.alias] = this.toResult(row[req.index], req.children)
+                    case 'list-lookup': {
+                        let rows = row[req.index]
+                        if (rows != null) {
+                            rec[req.alias] = this.toResult(row[req.index], req.children)
+                        } else {
+                            rec[req.alias] = []
```
tmcgroul, line 493: I think that query builder shouldn't specify default values but i haven't found a better place

Another way is to make default values explicit and extend FieldRequest interface with an optional field for default value
Or just iterate over data returned from querybuilder and search for null values which should be replaced by default value

But both variants are quite ugly

tmcgroul: @eldargab

eldargab: It's perfectly fine.

Please, reverse `rows != null` to `rows == null`

eldargab (changes requested):
Let's split this PR in two.

First to master branch with a new JSON scalar, as it is useful regardless of new release. Second to this branch with graphql schema for archive.

Apart from that, please provide all necessary tests [here](https://github.com/subsquid/squid/blob/2bdeb1b0c4c2674e4cb8567d48052b42dc251d88/openreader/src/test/scalars.test.ts#L1) demonstrating that new scalar works as intended.

openreader/src/gql/scalars/JSON.ts
```diff
@@ -0,0 +1,7 @@
+import {GraphQLScalarType} from "graphql"
+
+
+export const JSONScalar = new GraphQLScalarType({
+    name: 'JSON',
+    description: 'A scalar that can represent any JSON value',
```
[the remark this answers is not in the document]
tmcgroul: Yes, graphql lib does it fine by itself

PR #37, opened 2022-05-04
add support json scalar type to openreader

tmcgroul:


tmcgroul:
@eldargab postgres can sort by jsonb column but cockroachdb can't
according to it `orderBy: {json_field_ASC}` will work for postgres but will fail for cockroach
should i handle it somehow?

eldargab:
Lets omit `orderBy` for json scalars

tmcgroul:
done!
but it is not type safety... another way is to add property to scalars which will enable or desable orderBy

PR #39, opened 2022-05-08
Add storage batch calls

belopash:


substrate-processor/src/chain.ts
```diff
@@ -147,28 +147,66 @@ export class Chain {
...
+                assert(item.keys.length === keys.length)
+                query[i] = storageHash + this.getKeysHash(item, ...keys)
+            } else {
+                query[i] = storageHash + this.getKeysHash(item, keys)
+            }
+        }
+        let res = await this.client.call('state_queryStorageAt', [query, blockHash])
+        let changes: [string, string][] = res[0].changes
+        let results = new Array(query.length)
+        if (changes.length === results.length) {
```
belopash, line 171: Not sure how much time this will save

substrate-processor/src/chain.ts
```diff
@@ -147,28 +147,66 @@ export class Chain {
...
+            for (let i = 0; i < results.length; i++) {
+                results[i] = this.decodeStorageValue(item, changes[i][1])
+            }
+        } else {
+            for (let i = 0; i < results.length; i++) {
+                let value = changes.find(change => change[0] === query[i])?.[1]
```
belopash, line 177: `state_queryStorageAt` returns values whithout dups and nulls, so it is necessary to make it consistent with keys

PR #42, opened 2022-05-19
add contracts pallet support to substrate-processor

tmcgroul:


test/wasm-contracts/src/processor.ts
```diff
@@ -0,0 +1,40 @@
...
+    archive: assertNotNull(process.env.ARCHIVE)
+})
+
+
+processor.addContractsEventHandler(
+    "0xde8f2aa373e804aa57dd18b16de1249477375e205ec6c8a8bb129a5c4067a81c",
+    async ctx => {
+        let { event, block } = ctx.substrate;
+
+        let abi = new Abi(flipperMetadata)
```
dzhelezov, line 24: this should be a const and reused? creating Abi is probably very time consuming

test/wasm-contracts/src/processor.ts
```diff
@@ -0,0 +1,40 @@
...
+        let bytes = Buffer.from(ctx.data.slice(2), 'hex')
+        let decoded = abi.decodeEvent(bytes)
```
dzhelezov, line 26: Feeling very uneasy about the dependency of polkadot.js here as even they admit it as 'unstable': https://github.com/polkadot-js/api/blob/8649c17dd7055a82a719a3d9f489df6d406c4fd1/packages/api-contract/src/Abi/index.ts#L121

test/wasm-contracts/src/processor.ts
```diff
@@ -0,0 +1,40 @@
...
+const processor = new SubstrateProcessor(new TypeormDatabase('wasm_contracts'))
+
+
+processor.setDataSource({
+    chain: assertNotNull(process.env.CHAIN_NODE),
+    archive: assertNotNull(process.env.ARCHIVE)
+})
+
+
+processor.addContractsEventHandler(
```
dzhelezov, line 19: as mentioned above, to my ear `processor.addWasmEventHandler` sounds more natural and self-explanatory

dzhelezov:
Superb job so far!
Stuff to discuss/thing about:
- Naming (Contracts -> Wasm)
- Polkadot.js dependency and if we can live with it
- Call handlers (so far only event handlers are supported)

test/wasm-contracts/src/processor.ts
tmcgroul: i don't mind:)
when i chose addContractsEventHandler i was thinking about another pallets which support wasm
gear-tech for example
probably they will have different handler signature

test/wasm-contracts/src/processor.ts
tmcgroul: it's just a test project and processor doesn't require this lib
a decision which abi lib to use is up to mapping developers

PR #43, opened 2022-05-19
fix schema name

tmcgroul:
The queries will fail if `this.statusSchema` contains hypen

### DIFF
--- a/stores/typeorm-store/src/database.ts
+++ b/stores/typeorm-store/src/database.ts
@@ -21,6 +21,7 @@ export class TypeormDatabase {
     private lastCommitted = -1
 
     constructor(name: string, options?: TypeormDatabaseOptions) {
+        assert(/^[a-z][a-z0-9_]*$/.test(name), `name '${name}' is invalid`)
         this.statusSchema = `${name}_status`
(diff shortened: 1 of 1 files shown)


stores/typeorm-store/src/database.ts
```diff
@@ -33,18 +33,18 @@ export class TypeormDatabase {
         let con = await createConnection(cfg)
         try {
             let height = await con.transaction('SERIALIZABLE', async em => {
-                await em.query(`CREATE SCHEMA IF NOT EXISTS ${this.statusSchema}`)
+                await em.query(`CREATE SCHEMA IF NOT EXISTS "${this.statusSchema}"`)
                 await em.query(`
-                    CREATE TABLE IF NOT EXISTS ${this.statusSchema}.status (
+                    CREATE TABLE IF NOT EXISTS "${this.statusSchema}.status" (
```
eldargab, line 38: Here, it should be `"${this.statusSchema}".status`.

Also, I think we should restrict `statusSchema` to match `/^[a-z][a-z0-9_]*$/`.

eldargab (changes requested):

PR #49, opened 2022-06-12
implement wasm-typegen and wasm-decoder packages

tmcgroul:


tmcgroul:
@eldargab which name is better? wasm-typegen or contracts-typegen?

PR #52, opened 2022-06-20
New squid api

mo4islona:


cli/src/rest-client/request.ts
```diff
@@ -32,23 +32,29 @@ export async function request(
             console.log(chalk.dim(JSON.stringify(responseBody, null ,2)));
```
eldargab, line 32: `if (responseBody)` ?

The whole file looks suspiciously similar to `api.ts`

cli/src/api/api.ts
```diff
@@ -0,0 +1,73 @@
...
+    if (body) {
+      console.log(chalk.dim(JSON.stringify(body, null, 2)));
+    }
+  }
+
+  switch (response.status) {
+    case 200:
+      return { body };
+    default:
+      throw new ApiError(response.status, body as any);
```
eldargab, line 71: The fact that body can't be a stream here is not obvious from surrounding code.

I'd suggest to handle log streaming in a separate function.

eldargab (approved):
We can merge at the moment, but, please, have a look at comments.

cli/src/rest-client/request.ts
mo4islona: It is deprecated now. Will be removed soon

cli/src/config/utils.ts
```diff
@@ -6,46 +6,69 @@ import {
...
 
-export const configDirectory = resolve(homedir(), '.hydra-cli');
+export const configDirectory = resolve(homedir(), '.subsquid-cli');
 export const configFilePath = resolve(configDirectory, 'config.json');
 
 export const defaultApiUrl = 'https://saas.infra.gc.subsquid.io/api';
 
+export type Config = {
+    apiUrl: string
+    credentials: string
```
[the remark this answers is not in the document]
mo4islona: This will be refactored soon.
Credentials is valid only against that API url. 

We we eventually have multiple contexts
```ts
{
  currentContext: "dev",
  contexts : {
   {
     name: "production"
     apiUrl: "https://subsquid.io",
     token: "TOKEN_A"
    },
    {
     name: "dev",
     apiUrl: "https://dev.subsquid.io",
     token: "TOKEN_B"
    }  
  ]
}
```

cli/src/api/api.ts
mo4islona: We just need a type checking here

I`d suggest to explicitly set expecting response type
```
const { data } = await axios.get('https://stream.example.com', {
    headers: {Authorization: `Bearer ${token}`, 
    responseType: 'stream'
});

// data is stream
```

cli/src/config/utils.ts
```diff
@@ -6,46 +6,69 @@ import {
...
-        {
-            flag: 'w',
-            encoding: 'utf8',
-        }
+      configFilePath,
+      JSON.stringify(data),
+      {
+          flag: 'w',
+          encoding: 'utf8',
+      }
```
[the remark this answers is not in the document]
mo4islona: Maybe
I just do not touched what worked

PR #59, opened 2022-06-26
publish images to subsquid org

tmcgroul:
@eldargab should i also push images as image-name:major and image-name:latest?
and probably tags shoud be taken from package.json

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -20,20 +20,20 @@ jobs:
       - name: Docker login
         uses: docker/login-action@v1
         with:
-          username: eldar
-          password: ${{ secrets.EG_DOCKER_PASSWORD }}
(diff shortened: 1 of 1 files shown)


eldargab:
Hey, docker publishing is almost ready - https://github.com/subsquid/squid/blob/5b82c683b2253e78af15d3518912853fd9ed4b12/ops/docker-publish.sh#L7 :)

But, substrate gateway should follow the same tagging scheme.

tmcgroul:
> Hey, docker publishing is almost ready -
> 
> https://github.com/subsquid/squid/blob/5b82c683b2253e78af15d3518912853fd9ed4b12/ops/docker-publish.sh#L7
> :)
> 
> But, substrate gateway should follow the same tagging scheme.

Nice! Will fix the gateway later

PR #64, opened 2022-06-30
save contracts address in call table

tmcgroul:


eldargab (changes requested):
Please add a change file 

```
rush change -b origin/master
```

PR #72, opened 2022-07-13
make evmLog contract address case insensitive

belopash:


eldargab (changes requested):
Hey, address normalisation should be done at `.addEvmLogHandler()` and `.addEvmLog()`
level. Otherwise, handler based processor will not work.

belopash:
@eldargab should we do the same with contract addresses for `.addContractsContractEmittedHandler()` and `addContractsContractEmitted()`?

eldargab:
> should we do the same with contract addresses for `.addContractsContractEmittedHandler()` and `addContractsContractEmitted()`?

You mean`.toLowerCase()`? Seems like we should.

eldargab:
Hey, @RaekwonIII and @dzhelezov  !

For batch processing the above normalisation will not help, because an explicit dispatch on a contract address will be often required.

```ts
if (item.event.args.contract == '0xaA...') {
else {
}
```

Docs should clearly advise to work only with decoded raw addresses in a lower cased hex form.

PR #73, opened 2022-07-13
store gear program_id in the event table

tmcgroul:


tmcgroul:
@eldargab
I have two kind of events: MessageEnqueued and UserMessageSent which can have the same contract
Should i change current index `CREATE INDEX IDX_event__contract__block ON event(contract, block_id);` to:
`CREATE INDEX IDX_event__contract__block__name ON event(contract, block_id, name);`? Or just filter it in the gateway?

PR #80, opened 2022-07-25
add `X-SQUID-ID` request header to processor requests

belopash:


mo4islona (approved):

eldargab (changes requested):
Please add the change files and change the HTTP headers type to `Record<string, string>`

substrate-processor/src/processor/runner.ts
```diff
@@ -103,6 +103,9 @@ export class Runner<S, R extends BatchRequest> {
             }, 'request')
 
             let response = await graphqlRequest({
+                headers: {
+                    'x-squid-id': process.env.SQUID_ID || 'unknown',
```
dzhelezov, line 107: From @mo4islona:
Let's use `gen-{randomString(10)}` instead of `unknown`

dzhelezov (changes requested):

dzhelezov (approved):

mo4islona:
> It will be regenerated on every request?
> 
> We should preserve it between restarts

Oh, I see. Def decorator caches the method

PR #88, opened 2022-08-02
Feat/improve deploy flow

mo4islona:


eldargab:
@mo4islona , the correct target branch is `v0`. `release/v0` is supposed to be updated only by actions.

PR #93, opened 2022-08-13
add ethereum transactions to processor

belopash:


belopash:
I decided not to implement `decodeTxInput`

substrate-processor/src/processor/handlerProcessor.ts
```diff
@@ -913,6 +968,19 @@ class HandlerRunner<S> extends Runner<S, DataHandlers>{
                             log.debug('end')
                         }
                     }
+                    for (let handler of this.getEthereumTranscationHandlers(handlers, item.call)) {
+                        if (item.call.success || handler.triggerForFailedCalls) {
```
eldargab, line 972: Hey, the problem here is that successful execution of substrate call doesn't imply successful evm transaction.

https://github.com/paritytech/frontier/blob/ef75c42e9eb65319fd09ae589e3d965633b2ef51/frame/ethereum/src/lib.rs#L597

So, the processing of EVM transactions should be based on `Ethereum.Executed` event.

eldargab (changes requested):

substrate-processor/src/processor/handlerProcessor.ts
belopash: Is it possible to do it on processor side or should I fix it in ingester?

tmcgroul: I came up with two options:
1. A mapping developer should request 'Ethereum.Executed' event itself and check if a transaction is successful
2. The substrate-gateway should annotate `call.evmTxSuccess` field like it does for `event.evmTxHash` (this option rely on network runtime which is error prone)

eldargab:
Closing in favour of #98 . Let's continue work and discussion there.

PR #94, opened 2022-08-14
improve xcm types

belopash:


belopash:
@eldargab I guess the only problem with types was typos in V2, thay you already have fixed in prev commits. I added `Versioned*` types to index.ts and prefixed all other types with Xcm.
How can I generate interfaces?

eldargab:
@belopash , to generate interfaces run

```bash
node ops/xcm-typegen/lib/main.js
```

PR #95, opened 2022-08-15
New openreader

eldargab:


openreader/src/db.ts
```diff
@@ -1,83 +1,72 @@
...
+export class PoolOpenreaderContext implements OpenreaderContext {
+    private tx: LazyTransaction<Database>
+    private subscriptionPool: Pool
 
-    constructor(private pool: Pool) {}
+    constructor(
+        public readonly dialect: Dialect,
+        pool: Pool,
+        subscriptionPool?: Pool,
+        private subscriptionPollInterval: number = 1000
```
dzhelezov, line 34: It's better to set poll interval to 5 seconds since the average block time is about 6 seconds.

dzhelezov (approved):
LGTM modulo the default poll interval (1sec -> 5 sec). Need a bit more time to fully digest everything, but don't want to block the release.

PR #98, opened 2022-08-20
Ethereum transactions

eldargab:


eldargab:
Hey, @belopash , @tmcgroul !

Shall we support `sighash` and wildcard contract filters right away?

belopash:
It would be nice

> Hey, @belopash , @tmcgroul !
> 
> Shall we support `sighash` and wildcard contract filters right away?

substrate-processor/src/interfaces/gateway.ts
```diff
@@ -9,7 +9,7 @@ export interface BatchRequest {
     events?: ObjectRequest[]
     calls?: ObjectRequest[]
     evmLogs?: EvmLogRequest[]
-    ethereumTransactions?: EthereumTransactionRequest[]
+    ethereumTransactCalls?: EthereumTransactRequest[]
```
belopash: I think we should stick to one name `ethereumTransactions` or `ethereumTransactCalls`, overwise there can be confusion due different names in different parts of code

belopash: oh, was fixed in commite above

tmcgroul:
> Hey, @belopash , @tmcgroul !
> 
> Shall we support `sighash` and wildcard contract filters right away?
What if we add `sighash` support in a future release? Will we have to resync archive?

eldargab:
No, `sighash` is already archived by this PR.

tmcgroul:
In that case let's support `sighash`. I'll update the gateway

tmcgroul:
how about such interface:
`ethereumTransactions: [{contract: "*", selector: "0x095ea7b3"}]`?

eldargab:
I'm already putting `sighash` everywhere. Any reason why `selector` is better?

tmcgroul:
Only because i've already put `selector` in acala handlers.
Ok let it be sighash

belopash:
Has moonriver archive been migrated to new release? If yes, I think it's better to merge two eth test projects

PR #101, opened 2022-08-23
add `stateRoot` and `extrinsicsRoot` fields to processor

belopash:


eldargab (commented):
All good, except, we can't merge it at least until all our public archives will be migrated to a new version.

PR #104, opened 2022-08-25
Limits for openreader

eldargab:
This PR adds 2 new options to `graphql-server` and `openreader`.

* `--max-response-size <nodes>` - maximum number of JSON nodes in GQL response
* `--max-root-fields <count>` - maximum number of top level fields in a query

The reason for 2 options for basically the same kind of limit is due to our current implementation details. Each access to a top level field implies database request. Hence, queries

```gql
query {
   i1: itemById(id: "1") { id }
   # ... 
   i100: itemById(id: "100") { id }
}
```

and 

```gql
query {
   items(limit: 100) {
      id
   }
}
```

being similar in size, have completely different complexities (not regulated by other means).

There are also 2 new schema directives, that help with response size estimation:

* `@cardinality(value: Int!)` - applicable to entities and list relations
* `@byteWeight(value: Float!)` - applicable to scalars and typed JSON fields on entities

Have a look at `substrate-explorer` to see example usage and to play with queries.

### DIFF
--- a/substrate-explorer/schema.graphql
+++ b/substrate-explorer/schema.graphql
@@ -1,10 +1,10 @@
-type Metadata @entity {
+type Metadata @entity @cardinality(value: 100) {
     id: ID!
     specName: String!
     specVersion: Int
     blockHeight: Int!
     blockHash: String!
-    hex: String!
+    hex: String! @byteWeight(value: 1000.0)
 }
 
 type Block @entity {
@@ -17,9 +17,9 @@ type Block @entity {
     timestamp: BigInt!
     spec: Metadata!
     validator: String
-    events: [Event!]! @derivedFrom(field: "block")
-    calls: [Call!]! @derivedFrom(field: "block")
-    extrinsics: [Extrinsic!]! @derivedFrom(field: "block")
+    events: [Event!]! @derivedFrom(field: "block") @cardinality(value: 1000)
+    calls: [Call!]! @derivedFrom(field: "block") @cardinality(value: 1000)
+    extrinsics: [Extrinsic!]! @derivedFrom(field: "block") @cardinality(value: 1000)
 }
 
 type Call @entity {
@@ -31,7 +31,7 @@ type Call @entity {
     error: JSON
     origin: JSON
     name: String!
-    args: JSON
+    args: JSON @byteWeight(value: 5.0)
     pos: Int!
 }
 
@@ -43,7 +43,7 @@ type Event @entity {
     extrinsic: Extrinsic
     call: Call
     name: String!
-    args: JSON
+    args: JSON @byteWeight(value: 5.0)
     pos: Int!
 }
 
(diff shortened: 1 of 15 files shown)


belopash:
I set `--max-response-size 2`, but when I try to query
```
query MyQuery {
  accounts(limit: 2) {
    id
  }
}
```
it returns `"message": "response might exceed the size limit"`

Is root node also counted?

belopash:
And could you please explain how new derictives relate with options?

belopash:
And could you please explain what `cardinality` is for?

belopash:
And why this query also returns error? `--max-response-size 10`
```
query MyQuery {
  historicalBalances(limit: 1) {
    id
    account {
      balance
      historicalBalances(limit: 2) {
        id
        account {
          id
        }
      }
    }
  }
}
```
there are only 8 nodes in response (including root). Or arrays are also counted as roots?

eldargab:
>Or arrays are also counted as nodes?

Yes, arrays and objects are counted as nodes by themselves. Node counting roughly approximates the amount of memory query can potentially use, but it is not something particularly definite and precise.

 >And could you please explain what `cardinality` is for?

`@cardinality` is a way to specify typical number of elements in relation or table. For example,

```gql
type Order @entity {
   id: ID!
  name: String!
  # We rarely have more than 10 items, convey this information via @cardinality
  items: [Item!]! @derivedFrom(field: "order") @cardinality(value: 10) 
}

type Item @entity {
   id: ID!
  name: String!
  order: Order!
}
```

```gql
query {
  orderById(id: "1") {
    name
    items { # Although there is no limit, it is assumed to be 10
       id 
       name
    }
  }
}
```

substrate-explorer/schema.graphql
```diff
@@ -1,10 +1,10 @@
...
     id: ID!
     specName: String!
     specVersion: Int
     blockHeight: Int!
     blockHash: String!
-    hex: String!
+    hex: String! @byteWeight(value: 1000.0)
 }
 
 type Block @entity {
```
dzhelezov, line 10: We should add `@cardinality(value: 1_000_000)` here to enforce a limit?

substrate-explorer/schema.graphql
```diff
@@ -17,9 +17,9 @@ type Block @entity {
...
     validator: String
-    events: [Event!]! @derivedFrom(field: "block")
-    calls: [Call!]! @derivedFrom(field: "block")
-    extrinsics: [Extrinsic!]! @derivedFrom(field: "block")
+    events: [Event!]! @derivedFrom(field: "block") @cardinality(value: 1000)
+    calls: [Call!]! @derivedFrom(field: "block") @cardinality(value: 1000)
+    extrinsics: [Extrinsic!]! @derivedFrom(field: "block") @cardinality(value: 1000)
 }
 
 type Call @entity {
```
dzhelezov, line 25: Same here, `@cardinality(value: 1_000_000)`

dzhelezov (approved):
LGTM -- put need to review to the Explorer schema (add top-level cardinalities with large numbers to make sure limits are enforced)

PR #110, opened 2022-09-09
Add logic to ignore the `tip` field if it's not a numeric type

ken-centrality:
Close #108 

## Changes
- Ignore the `tip` field if it's non-numeric
- Log a debug message to explain why

PR #111, opened 2022-09-09
Add `Promise.race` to avoid ROLLBACK query getting stuck

### LINKED ISSUE #109: "ROLLBACK" query got stuck
**Describe the bug**
At the back of #108, I found another issue where the process got stuck [here](https://github.com/subsquid/squid/blob/master/substrate-ingest/src/sink.ts#L290)

**To Reproduce**
1. Follow the step as of #108

**Expected behavior**
Wait for "ROLLBACK" query, if it's longer than X time, ignore

ken-centrality:
Closes #109

### DIFF
--- a/substrate-ingest/src/sink.ts
+++ b/substrate-ingest/src/sink.ts
@@ -287,7 +287,10 @@ export class PostgresSink implements Sink {
             await this.db.query('COMMIT')
             return result
         } catch(e: any) {
-            await this.db.query('ROLLBACK').catch(() => {})
+            await Promise.race([
+                this.db.query('ROLLBACK').catch(() => {}),
+                new Promise((resolve) => setTimeout(resolve, 5000))
+            ])
             throw e
         }
     }


ken-centrality:
any chance this and #110 get reviewed and merge?

eldargab:
The point of `ROLLBACK` query is to keep the connection in a clean state. That doesn't happen when you time out rollback query like this PR suggests. You could omit rollback query entirely with the same effect.

As of why `ROLLBACK` hangs - this is a [node-postgres](https://github.com/brianc/node-postgres) issue which doesn't properly handle serialization errors and leaves the client state corrupted after those.

ken-centrality:
@eldargab I agreed,  it's an issue on `node-postgres`, but don't you think it's better to throw and terminate the process if the ROLLBACK query takes a long time? Or perhaps a WARN so at least the dev knows what's going on?

PR #112, opened 2022-09-09
Local cache for openreader

dzhelezov:


eldargab:
There are problems.

First, after having a thorough look at implementation of Apollo request processing pipeline, it seems very dubious to me. I wouldn't wonder if it will appear, that Apollo server itself is significantly responsible for slow result serialisation performance. That's why I want to get rid from Apollo in near future.

Second, the caching is done on a per-field basis. This can easily lead to data consistency issues. Of course, causality anomalies can happen for per-query caching as well, but having discrepancies in the same response seems even more dangerous to me.

dzhelezov:
> There are problems.
> 
> First, after having a thorough look at implementation of Apollo request processing pipeline, it seems very dubious to me. I wouldn't wonder if it will appear, that Apollo server itself is significantly responsible for slow result serialisation performance. That's why I want to get rid from Apollo in near future.

I did some tests with fetching 10k accounts from the balances example. It was consistently under 50ms when cached, and 150+ms when the cached expired. I believe the cache does give a significant performance improvement when

- the same client does polling with identical requests
- multiple clients request the same data (e.g. analytical queries of Kodadot or Moonsama, initial data of the marketplace page) 
- it greatly relieves the pressure on the API server's RAM as the response JSON's are almost instantly GC'd

We may replace Apollo at some point in the future, but I would like to test the existing solution and see how it performs on the existing squids. If the performance is acceptable, we can stick with it for a while.

> Second, the caching is done on a per-field basis. This can easily lead to data consistency issues. Of course, causality anomalies can happen for per-query caching as well, but having discrepancies in the same response seems even more dangerous to me.

The `responseCachePlugin` used here only caches full query responses: https://github.com/apollographql/apollo-server/tree/main/packages/apollo-server-plugin-response-cache 
This cache plugin is pretty dumb: https://github.com/apollographql/apollo-server/blob/40ed23fbb5dd620902d7c31bcc1e26e098990041/packages/apollo-server-plugin-response-cache/src/ApolloServerPluginResponseCache.ts#L195

I don't see how it could lead to any data inconsistencies here.

PR #113, opened 2022-09-12
retrieve validator from indexer gateway

acdibble:
Adds `validator` to the gateway query and passes though the value to the handlers.

### DIFF
--- a/substrate-processor/src/ingest.ts
+++ b/substrate-processor/src/ingest.ts
@@ -236,6 +236,7 @@ export class Ingest<R extends BatchRequest> {
                     q.line('specId')
                     q.line('stateRoot')
                     q.line('extrinsicsRoot')
+                    q.line('validator')
                 })
                 q.line('events')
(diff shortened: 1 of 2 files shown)


substrate-processor/src/ingest.ts
```diff
@@ -390,10 +391,10 @@ function tryMapGatewayBlock(block: gw.BatchBlock): BlockData {
 
     items.sort((a, b) => getPos(a) - getPos(b))
 
-    let {timestamp, ...hdr} = block.header
+    let {timestamp, validator, ...hdr} = block.header
 
     return {
-        header: {...hdr, timestamp: new Date(timestamp).valueOf()},
+        header: {...hdr, timestamp: new Date(timestamp).valueOf(), validator: validator ?? undefined},
```
acdibble, line 397: the gateway will return `string | null` for `validator`, so I've casted it to preserve backwards compatibility.

PR #119, opened 2022-09-22
intoduce `substrate-frontier-evm` package

belopash:


dzhelezov (approved):
LGTM modulo error messages

util/substrate-frontier-evm/src/transaction.ts
```diff
@@ -0,0 +1,155 @@
+import assert from 'assert'
+import {Call, ChainContext, Event} from './interfaces'
```
tmcgroul, line 2: `Event` is unused

util/substrate-frontier-evm/src/log.ts
```diff
@@ -0,0 +1,30 @@
+import {assert} from 'console'
```
tmcgroul, line 1: Shouldn't it be `import assert from 'assert'`?

tmcgroul (changes requested):

tmcgroul:
@belopash Also please add a README

PR #120, opened 2022-09-26
fix naming collision between Acala EVM and frontier

tmcgroul:


substrate-ingest/src/sink.ts
```diff
@@ -393,6 +395,10 @@ export class PostgresSink implements Sink {
             throw e
         }
     }
+
+    private isAcalaEvmExecuted(event: Event) {
+        return 'contract' in event.args && 'logs' in event.args
```
tmcgroul, line 400: there are more than two fields but i believe it's enough
https://acala.subscan.io/extrinsic/1896016-2?event=1896016-45

dzhelezov (approved):

PR #136, opened 2022-12-14
Fix ink-v3-schema: dispatchKey is string type

HyunggyuJang:
Like cell's key, dispatchKey is always string type.

### DIFF
--- a/substrate/ink-abi/src/metadata/ink-v3-schema.json
+++ b/substrate/ink-abi/src/metadata/ink-v3-schema.json
@@ -215,11 +215,7 @@
       "properties": {
         "dispatchKey": {
           "description": "The key where the discriminant is stored to dispatch the variants.",
-          "allOf": [
-            {
-              "$ref": "#/definitions/LayoutKey"
-            }
-          ]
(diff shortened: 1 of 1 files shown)


tmcgroul:
@HyunggyuJang could you please share details/examples to explain impact of your changes?
I took the json schema from some sort of official source - https://github.com/paritytech/ink/blob/master/crates/metadata/ink-v3-schema.json#L220 - is ink's schema invalid?

HyunggyuJang:
Yes, I think so. https://github.com/AstarNetwork/wasm-showcase-dapps/blob/main/uniswap-v2/contracts/pair/lib.rs is an example at my hand, and indeed all dispatchKey are string as far as know (at least for all contracts in that repo).

HyunggyuJang:
Just note that, with the up-to-date [schema file,](https://github.com/paritytech/ink/blob/master/crates/metadata/ink-v3-schema.json) the [one from your template](https://github.com/subsquid/squid-wasm-template/blob/master/src/abi/erc20.json) is also invalid.

PR #137, opened 2022-12-14
Fix corrupted message name if abi contains messages with path

HyunggyuJang:
For example, if we use openbrush's PSP22, it will generate
```typescript
export interface Message_PSP22::allowance {
  __kind: 'PSP22::allowance'
  owner: Uint8Array
  spender: Uint8Array
}
```
which is not valid naming in typescript. With this PR, it generates
```typescript
export interface Message_PSP22_allowance {
  __kind: 'PSP22_allowance'
  owner: Uint8Array
  spender: Uint8Array
}
```

### DIFF
--- a/substrate/ink-abi/src/abi-description.ts
+++ b/substrate/ink-abi/src/abi-description.ts
@@ -55,7 +55,7 @@ export class AbiDescription {
             kind: TypeKind.Variant,
             variants: list.map((msg, index) => {
                 return {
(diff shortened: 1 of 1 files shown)


belopash:
looks like this will be fixed in https://github.com/subsquid/squid-sdk/pull/135

tmcgroul:
@HyunggyuJang Thanks for your contribution! But you are a bit late...

PR #141, opened 2022-12-19
Add Request data from ExpressContext to graphql server's Context

Lezek123:
Add Request data from ExpressContext to graphql server's Context to allow retrieving useful request data, like client ip, HTTP headers inside custom resolvers, allowing the squid developer to implement authentication, rate limiting etc. for the custom graphql endpoints.

A simple example usage: https://github.com/Lezek123/orion/blob/orion-v2/src/server-extension/resolvers/VideosResolver/index.ts#L47

### DIFF
--- a/common/changes/@subsquid/graphql-server/lw-add-request-to-context_2022-12-19-11-08.json
+++ b/common/changes/@subsquid/graphql-server/lw-add-request-to-context_2022-12-19-11-08.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@subsquid/graphql-server",
+      "comment": "Add Request data from ExpressContext to graphql server's Context object",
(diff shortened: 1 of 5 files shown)


eldargab:
There is "check extension" in graphql-server - https://github.com/subsquid/squid-sdk/blob/ab1cae1eb6c9c4a34760f4fea09dff45c2f4065b/graphql/graphql-server/src/check.ts#L41

Graphql context is already available at that point, you just need to pass it to the check function. 

That will allow you to patch it from custom extension or to do some rate-limiting/validations before hitting the main graphql machinery.

eldargab:
Graphql context is now exposed to check extension - https://github.com/subsquid/squid-sdk/commit/063887ab9ce2ce7327fab581faa9f39600b122d2

PR #142, opened 2022-12-22
add upsert alias for store.save()

dzhelezov:


belopash:
Why do we need this?

dzhelezov:
At least two devs were using `save()` + `insert()`. So the name `save()` seems to be misleading, `upsert()` is more accurate.

belopash:
Hmm, maybe it's better to make "save" an alias instead? It will just be more logical

dzhelezov:
I am neutral with respect to that, let @eldargab decide and merge :)

eldargab:
Yes, it's better for `.save()` to be an alias. Because that way its meaning will become easier to discover:

1. jump to definition
2. immediately see, that `.save() == .upsert()`.

Versus

1. jump to definition
2. read a long description text

PR #152, opened 2023-01-24
Ink abi fix

### LINKED ISSUE #151: ink! smart contract without events should have any type
**Describe the bug**

I have a smart contract that does not emit any events
- https://github.com/swanky-dapps/marketplace/issues/9

the problem is when I do `npx squid-ink-typegen` (+ args ofc)

it generates content like:

```ts
export function decodeEvent(hex: string): Event {
    return _abi.decodeEvent(hex)
}

export type Event = never
```

Now it it impossible to build the squid 

```
src/abi/marketplace.ts:1412:5 - error TS2322: Type 'any' is not assignable to type 'never'.

1412     return _abi.decodeEvent(hex)
         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~


Found 1 error in src/abi/marketplace.ts:1412
```

1. FIX is to return any if there is no event present 

```ts
export type Event = any
```

or
2. do not generate `decodeEvent` function

tmcgroul:
fix #151

### DIFF
--- a/common/changes/@subsquid/ink-abi/ink-abi-fix_2023-01-24-11-16.json
+++ b/common/changes/@subsquid/ink-abi/ink-abi-fix_2023-01-24-11-16.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@subsquid/ink-abi",
+      "comment": "fix type error for contracts without events",
+      "type": "patch"
+    }
+  ],
+  "packageName": "@subsquid/ink-abi"
+}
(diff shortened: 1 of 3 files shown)


belopash:
Better not to generate `decodeEvent` if there is no events or use an old approach

belopash:
Now it anyway returns `never` type, doesn't it?

tmcgroul:
yes it still generates a function annotated with `never` but now it doesn't fail compilation
i don't wanna mess the code with conditions:)

Issue #166, opened 2023-03-09
Error encountered with "wrapAccessError" message while decoding ABI for event with property "owner"


Cupnfish:
**Describe the bug**
I encountered an error in my program named "wrapAccessError". The error message displayed "Error: deferred error during ABI decoding triggered accessing property "owner"". This error confuses me as the same event has been processed multiple times before, but this one failed.
**To Reproduce**

    Use the following data for testing:
    id: 0005648711-000174-018c3
    address: 0x314159265dd8dbb310642f98f50c066173c1259b
    data: 0x6330363834636235336331363831343865616130313363333864316330663339
    index: 174
    topics: ["0xce0457fe73731f824cc272376169235128c118b49d344817417c6d108d155e82","0x8352da3d0ebe15fd4bd7def280458b52cdb17c9c50ef26bed05f77a09a37033d","0x6630353636393265383962393531363132383865306565373939353535636238"]
    transactionIndex: 102

    Decode the data structure:
    [node: string, label: string, owner: string] & {
    node: string;
    label: string;
    owner: string;
    }

    The error "wrapAccessError" occurs with the message "Error: deferred error during ABI decoding triggered accessing property "owner""

**Expected behavior**
I expected the event to be processed successfully like the previous times.

**Screenshots**
N/A

belopash:
Hello, we use ethersV5 for evm data decoding.
[Here](https://github.com/ethers-io/ethers.js/issues/891#issuecomment-932985011) dev says that error means that event has invalid UTF8 string.

Cupnfish:
That seems to be the case, so how do I go about decoding the data? It seems to require a change in squid's internal code? @belopash

belopash:
> That seems to be the case, so how do I go about decoding the data? It seems to require a change in squid's internal code? @belopash 

I think you can change type in abi from `string` to `bytes` and then manually convert it to utf8 string

Cupnfish:
Here are the details of the error that occurred earlier:
```json
[...]

PR #168, opened 2023-03-21
migrate to ethers-v6

belopash:


Cupnfish:
May I ask if this pull request has the potential to address the issue reported in this bug report: https://github.com/subsquid/squid-sdk/issues/166?

belopash:
@Cupnfish actually I no, but maybe it will be fixed. As I said before, it's issue on ethers side, so we can only suggest workaround

PR #169, opened 2023-03-23
Fix for big-typed numeric arrays

abernatskiy:
Tested the BigInt arrays and found no issues. BigDecimals were failing to serialize properly, fixed that.


belopash:
Does reading work?

abernatskiy:
Didn't notice the issue with store reads. Will work on that next week.

PR #174, opened 2023-04-16
Allow _ prefix for type and prop

jasl:
Allow _ prefix seems secure.

My case is, on-chain, an NFT item full represent is `(CollectionId, ItemId)`, in squid, the below code will have a conflict

```
type NftExtendInfo @entity {
  id: ID!

  nftItem: NftItem! # It generates a varchar column nft_item_id to the database to store the foreign key, the format would like `${NftCollectionId}-${NftItemId}`
  nftItemId: Int! # This is the pure on-chain NFT item id
}
```

By introducing _ prop prefix, I would like to change it to

```
type NftExtendInfo @entity {
  _id: ID!

  _nftItem: NftItem!
  nftItemId: Int!
}
```

In addition, I also introducing _ prefix for type, I'm not use that, but I think it helpful for some cases

### DIFF
--- a/graphql/openreader/src/model.tools.ts
+++ b/graphql/openreader/src/model.tools.ts
@@ -137,8 +137,8 @@ export function validateModel(model: Model) {
 }
 
 
-const TYPE_NAME_REGEX = /^[A-Z][a-zA-Z0-9]*$/
-const PROP_NAME_REGEX = /^[a-z][a-zA-Z0-9]*$/
+const TYPE_NAME_REGEX = /^_?[A-Z][a-zA-Z0-9]*$/
+const PROP_NAME_REGEX = /^_?[a-z][a-zA-Z0-9]*$/
 
 
 export function validateNames(model: Model) {


jasl:
BTW, when I try run `make test` in local, `npm i` failed
```
npm ERR! code E404
npm ERR! 404 Not Found - GET https://registry.npmjs.org/gql-test-client - Not found
npm ERR! 404
npm ERR! 404  'gql-test-client@^0.0.0' is not in this registry.
npm ERR! 404
npm ERR! 404 Note that you can also install from a
npm ERR! 404 tarball, folder, http url, or git URL.
```

belopash:
`_property` and `_Property` will collide. I think correct regex will be `^_?[a-z][a-zA-Z0-9]*$`

jasl:
> `_property` and `_Property` will collide. I think correct regex will be `^_?[a-z][a-zA-Z0-9]*$`

Thank you, yours is better, updated

jasl:
Try patching local one, it seems update the regex is not enough, will fail in `validateLookups`

jasl:
hello?

jasl:
ping

eldargab:
Hey, sorry for not responding!

Actually `_` prefix were reserved for technical fields and that was used in a couple of places. 
We can eliminate this limitation, but that will require a bit more than merging PR and right now we don't have time for that.

jasl:
> Hey, sorry for not responding!
> 
> Actually `_` prefix were reserved for technical fields and that was used in a couple of places. We can eliminate this limitation, but that will require a bit more than merging PR and right now we don't have time for that.

Yeah, I hope you can put it in your backlog.

Although my example is invalid now, but our Solana and EVM contracts have fields such as `id`, `type`, or `tokenID`, I want columns in Squid to keep the same names as our contracts to reduce misleading, and Squid's ID fields is less useful in dAPP in my experience

jasl:
typeORM's convention of association columns is difficult to debug. I remember this PR because I encountered a naming collision, which took me an hour to figure out.

belopash:
Alternatively, you can use another graphql server, like https://www.graphile.org/postgraphile/ for example.

Usage example:
https://github.com/subsquid-labs/squid-postgraphile-example/

PR #178, opened 2023-05-28
typegen for storage layout

belopash:


belopash:
I think it needs tests.

test/eth-usdc-transfers/src/abi/usdc.ts
```diff
@@ -0,0 +1,116 @@
...
+    }
+}
+
+class Uint256Item extends StorageItem<bigint> {}
+
+class Uint8Item extends StorageItem<bigint> {}
+
+export const storage = {
+    _owner: new AddressItem(
+        layout, 't_address', 0n, 0
```
belopash, line 54: don't like that need to pass type again, but otherwise typegen become bulky

eldargab:
Unfortunately, this is not ready for release

PR #179, opened 2023-06-02
fix kusama type definitions

belopash:


belopash:
Maybe there are more type inconsistencies, since fixed type should probably affect other chains too

PR #190, opened 2023-07-20
Make OneToOne releations bi-directional

belopash:


eldargab:
What is the case which bothered you enough to make a PR? :)

belopash:
@eldargab missed the comment

```
type Account @entity {
    id: ID!
    publicKey: ID! @index
    stashOf: Staker @derivedFrom(field: "stash")
    controllerOf: Staker @derivedFrom(field: "controller")
}

type Staker @entity {
    id: ID!
    stash: Account! @unique
    controller: Account @unique

    ...
}
```
need to query account in code to get staker entity

belopash:
for now I patched it locally, so it's not urgent

prap1546:
good

PR #191, opened 2023-07-26
fix: NULL appears first when order By DESC

justraman:


eldargab:
Hey, this is a breaking change and it's non-obvious that it is always what users want!

I'm not against CLI flag enabling such behaviour, also would be nice to know how Hasura and postgraphile behave...

justraman:
Hey @eldargab , I see your point.  I just checked the Hasura and Postgraphile as mentioned
1. Postgraphile allows it by using an argument when initializing the engine `orderByNullsLast: true`
2. On the other hand, Hasura handles nullable fields differently by creating additional OrderBy options [desc_nulls_last, desc_nulls_first and default desc ](https://hasura.io/docs/latest/queries/postgres/sorting/)

Which one would you prefer?

eldargab:
Can we agree on 

```gql
query {
  list(orderBy: [foo_DESC_NULLS_LAST, bar_DESC, baz_ASC_NULLS_FIRST, qux_ASC]) {
    foo
    bar
  }
}
```

?

eldargab:
To elaborate:

With the flag option it is possible to have 2 similarly looking servers behaving completely differently. Also that approach is not flexible.

That leaves us with the need to extend the interface...

justraman:
> 

Absolutely!, I will work on it this week.

eldargab:
Great!

The very brief implementation guide will be

1. extend the [SortOrder](https://github.com/subsquid/squid-sdk/blob/ab1cae1eb6c9c4a34760f4fea09dff45c2f4065b/graphql/openreader/src/ir/args.ts#L85)
2. check and fix places where it is used :)

justraman:
@eldargab it's ready for review, i tried and tested locally.

graphql/openreader/src/opencrud/orderBy.ts
```diff
@@ -26,41 +26,45 @@ export function getOrderByMapping(model: Model, typeName: string): OpenCrud_Orde
...
-        switch(propType.kind) {
+        switch (propType.kind) {
             case 'scalar':
             case 'enum':
                 if (propType.name != 'JSON') {
-                    m.set(key + '_ASC', {[key]: 'ASC'})
-                    m.set(key + '_DESC', {[key]: 'DESC'})
+                    m.set(key + '_ASC', { [key]: 'ASC' })
+                    m.set(key + '_DESC', { [key]: 'DESC' })
+                    if (properties[key].nullable || nullable) {
```
eldargab, line 45: This check makes `nullable` -> `non-nullable` transition a breaking change. I would probably generate sort options regardless nullability.

graphql/openreader/src/ir/args.ts
```diff
@@ -82,4 +82,4 @@ export interface OrderBy {
 }
 
 
-export type SortOrder = "ASC" | "DESC"
+export type SortOrder = 'ASC' | 'DESC' | 'ASC_NULLS_FIRST' | 'DESC_NULLS_LAST'
```
eldargab, line 85: I'd use spaces here (`'ASC NULLS FIRST'`) to avoid introduction of regex replacement below

eldargab (changes requested):
Please include a change file as well

```bash
rush change -b origin/master
```

PR #199, opened 2023-09-18
New environments DB_SCHEMA and DB_SYNCHRONIZE

smaknsk:
This pull request adds DB_SCHEMA to make it convenient to share the same database for the application and the squid.
In some cases it is useful to use DB_SYNCHRONIZE instead of migrations.


smaknsk:
This pull request adds the stateSchema to the generator: https://github.com/subsquid/squid-gen/pull/49

eldargab:
As of `DB_SYNCHRONIZE` - this is never a good idea, especially with TypeORM broken migrations.

As of `DB_SCHEMA` - your changes only affect TypeORM settings, but we have direct `pg` connections as well. Also this is usually can be managed via `ALTER ROLE|USER name SET search_path TO schema_name`

PR #201, opened 2023-09-23
rework substrate typegen

belopash:
Typegen for storage is disabled. I'm not satisfied with current result, want to discuss


eldargab:
Sorry, can't accept it, due to lots of undiscussed, unnecessary changes.

PR #202, opened 2023-09-26
ignore STOP debug frame

tmcgroul:


evm/evm-processor/src/ds-rpc/mapping.ts
```diff
@@ -398,6 +398,9 @@ function* mapDebugFrame(transactionIndex: number, debugFrameResult: rpc.DebugFra
                     }
                 }
                 break
+            case 'STOP':
```
eldargab, line 401: Here we have a possibility of 

* gaps in trace indexes/addresses
* children with no parent

if `STOP` will turn out to be not the last frame in transaction or to have children.

Thus, it is better to ignore it on traversal.

eldargab (changes requested):

evm/evm-processor/src/ds-rpc/mapping.ts
eldargab: >i checked several STOP frames from three networks (base, optimism, zora) and all the time it was a single frame

Then, you have to make the assumption explicit by asserting it!

PR #215, opened 2023-11-22
Rename MessageEnqueued event to MessageQueued

osipov-mit:


dzhelezov:
We need to test agains ArrowSquid for Gear

eldargab:
Hey, @osipov-mit , is it for the whole mainnet or there are blocks with old names?

osipov-mit:
> Hey, @osipov-mit , is it for the whole mainnet or there are blocks with old names?

Yes it is for the whole mainnet

dzhelezov:
@abernatskiy we should update https://docs.subsquid.io/substrate-indexing/specialized/gear/ too

eldargab:
But after actual release

PR #216, opened 2023-11-25
tron native support

tmcgroul:


tron/tron-normalization/src/data.ts
```diff
@@ -0,0 +1,92 @@
...
+    cancelUnfreezeV2Amount?: Record<string, bigint>
+    result?: string
+    energyFee?: bigint
+    energyUsage?: bigint
+    energyUsageTotal?: bigint
+    netUsage?: bigint
+    netFee?: bigint
+    originEnergyUsage?: bigint
+    energyPenaltyTotal?: bigint
+    _transferContractOwner?: string
```
eldargab, line 57: I don't think, that introduction of this fields is justified. Their derivation is quite strait forward a can be done in-place.

eldargab (changes requested):

eldargab (approved):

PR #222, opened 2024-01-05
Add synchronize option for typeorm

smaknsk:
Hello!
I would like to use the **synchronize** option for development.
Also I have a case where I would like to use these options in my small project without creating and applying a migration.
And of course this is not recommended for production https://typeorm.io/faq#how-do-i-update-a-database-schema.


eldargab:
We had such request before, but I'm not in favour of it.

Will leave it open for other team members to decide. @dzhelezov , @mo4islona

eldargab:
Here is one example showing that it is not as benign as it looks.

Consider the case when both TypeORM based graphql server extensions and the processor is used. User can easily start both simultaneously leading to a race between migration procedures. At the same time we know, that TypeORM doesn't handle migration races properly.

smaknsk:
The default value is **false**. If someone sets this to **true**, they probably know what they're doing.
Squid-sdk is a framework and I think being able to control typeORM settings is a good thing :)

llllvvuu:
Generally people don't set this in prod. It is useful for dev, where too lazy to run migrations. I would be a user.

PR #224, opened 2024-01-15
add transaction items to log and trace requests

belopash:


evm/evm-processor/src/ds-rpc/filter.ts
```diff
@@ -140,6 +145,16 @@ export function filterBlock(block: Block, dataRequest: DataRequest): void {
             if (rel.transaction) {
                 include.addTransaction(log.transaction)
             }
+            if (rel.transactionLogs && log.transaction) {
```
belopash, line 148: archive also doesn't include transaction if it is not explicitly requested with `log.transaction`

eldargab: `.transactionLogs` and `.transactionTraces` don't imply transaction presence. Hence, log inclusion should not rely on `.transaction` property. 

`groupBy(logs, log => log.transactionIndex)` should be used instead.

evm/evm-processor/src/ds-rpc/filter.ts
```diff
@@ -140,6 +145,16 @@ export function filterBlock(block: Block, dataRequest: DataRequest): void {
...
+                for (let sibling of log.transaction.logs) {
+                    include.addLog(sibling)
+                }
+            }
+            if (rel.transactionTraces && log.transaction) {
+                for (let trace of log.transaction.traces) {
```
eldargab, line 154: Same as before, except this time presence of logs should be enforced [here](https://github.com/subsquid/squid-sdk/blob/16e7880a1a8428d173c6b50a3b4c9a653e47e93c/evm/evm-processor/src/ds-rpc/request.ts#L15) as well.

evm/evm-processor/src/ds-rpc/filter.ts
```diff
@@ -182,6 +197,11 @@ export function filterBlock(block: Block, dataRequest: DataRequest): void {
             if (rel.transaction) {
                 include.addTransaction(trace.transaction)
             }
+            if (rel.transactionLogs && trace.transaction) {
```
eldargab, line 200: Same as above.

eldargab (changes requested):

PR #231, opened 2024-02-14
transform bytes values to lowercase

belopash:


util/util-internal-validation/src/primitives.ts
```diff
@@ -129,12 +129,16 @@ export const SMALL_QTY: Validator<number, Bytes> = {
...
+        if (isBytes(value)) {
+            return value.toLowerCase() as Bytes
+        } else {
+            return new ValidationFailure(value, `{value} is not a hex encoded binary string`)
+        }
     },
     validate(value: unknown): ValidationFailure | undefined {
-        if (isBytes(value)) return
-        return new ValidationFailure(value, `{value} is not a hex encoded binary string`)
+        let i = this.cast(value)
```
eldargab, line 140: Here `this.cast()` performs extra `.toLowerCase()` work and potentially creates an extra object. This might be noticeable in a data intensive context.

That's why it is better to do `cast()` like it was originally - `this.validate(value) || (value as Bytes).toLowerCase()`

eldargab (changes requested):

eldargab:
@belopash , please push Bump and Release buttons.

PR #246, opened 2024-02-27
Fix typos

destinyae:
fix typo


tmcgroul:
@destinyae wouldn't you better to focus on existed problems instead of farming github activity?

destinyae:
> @destinyae wouldn't you better to focus on existed problems instead of farming github activity?

?????

PR #255, opened 2024-03-16
Evm ingestion reform

tmcgroul:
@eldargab it would be nice to get feedback on my changes. current state involves a lot of duplications and i don't really know how to avoid it


evm/evm-data/src/normalization/data.ts
tmcgroul, line 1: this file is almost identical copy of evm.ts from evm-processor. this one has more fields marked as optional

evm/evm-data/src/normalization/mapping.ts
tmcgroul, line 1: code from this file is very simiral to ds-rpc/mapping.ts file from evm-processor but i don't know how to unify them because they are related to different clients: this one for evm-ingest and writer and another one for processor's purposes

evm/evm-processor/src/ds-archive/schema.ts
```diff
@@ -1,29 +1,47 @@
...
+        hash: BYTES,
+        parentHash: BYTES,
+    })
 
     let Transaction = object({
-        hash: fields.transaction?.hash ? BYTES : undefined,
-        ...getTxProps(fields.transaction, true),
-        sighash: fields.transaction?.sighash ? withDefault('0x', BYTES) : undefined,
-        ...getTxReceiptProps(fields.transaction, true)
+        ...project(fields.transaction, {
```
tmcgroul, line 32: even though we have full schema defined we need to merge them and override fields that have different data types in archive. which is kind of ugly...

util/util-internal-validation/src/dsl.ts
```diff
@@ -17,15 +17,8 @@ import {GetCastType, GetSrcType, Validator} from './interface'
...
-): Validator<GetPropsCast<Props>, GetPropsSrc<Props>> {
-    let presentProps: Record<string, Validator<any>> = {}
-    for (let key in props) {
-        let v = props[key]
-        if (v) {
-            presentProps[key] = v
-        }
-    }
-    return new ObjectValidator(presentProps) as any
+): ObjectValidator<Props> {
```
tmcgroul, line 20: idk how to tell to ts compiler that this function has to return the same type that is passed to the function but without undefined values 🤷‍♂️

PR #258, opened 2024-03-21
Introduce evm-codec package with abi encoding/decoding

vanruch:


eldargab (commented):
I'd like to suggest the following codec structure - https://github.com/subsquid/squid-sdk/tree/evm-utils-feedback/evm/evm-codec-2

It is

* Free from any extra allocations (explicit or implicit)
* Clearly separates tokenisation (`Src` / `Sink`) from parsing and walking making things conceptually simpler. 

I haven't tested, but it should be significantly faster, especially for encoding, as `Buffer.alloc()`, I believe, is allocated outside of GC managed heap and must do some `malloc` style business, which is not nearly as cheep as regular JS allocation.

evm/evm-codec/src/codec.ts
```diff
@@ -0,0 +1,20 @@
...
+import type { Src } from "./src";
+import type { Pretty } from "./utils";
+
+export const WORD_SIZE = 32;
+
+export interface Codec<T> {
+  encode(sink: Sink, val: T): void;
+  decode(src: Src): T;
+  isDynamic: boolean;
+  slotsCount?: number;
```
belopash, line 11: why is it needed?

evm/evm-codec/src/codecs/array.ts
```diff
@@ -0,0 +1,85 @@
...
+    for (let i = 0; i < val.length; i++) {
+      val[i] = this.item.decode(tmpSrc);
+    }
+    src.jumpBack();
+    return val;
+  }
+}
+
+export class FixedArrayCodec<const T> implements Codec<readonly T[]> {
+  public isDynamic: boolean;
```
belopash, line 36: How "fixed" array can be "dynamic"? What does it mean? Why not to make separate struct?

evm/evm-codec/src/contract-base.ts
```diff
@@ -0,0 +1,59 @@
...
+export interface BlockContext {
+  _chain: Chain;
+  block: Block;
+}
+
+export interface Block {
+  height: number;
+}
+
+export class ContractBase {
```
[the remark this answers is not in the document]
vanruch: I'll extract function for performing eth calls, so that it can be oferridden if necessary

evm/evm-codec/src/codec.ts
vanruch: We need it to calculate the offsets for the dynamic types encoding. If it's not present, we just assume it takes 1 slot

evm/evm-codec/src/codecs/array.ts
vanruch: https://docs.soliditylang.org/en/latest/abi-spec.html#formal-specification-of-the-encoding

A type string[2], for example would be considered dynamic. Same as tuple (int,int,string). From the user's POV it's still a fixed array size, only the parsing logic is different, so imo it should stay as a same struct

vanruch: I'll rename it to FixedSizeArray to avoid confusion

PR #261, opened 2024-03-26
Migrate EVM typegen to new codec

vanruch:


evm/evm-typegen/src/pretty-out-dir.ts
```diff
@@ -0,0 +1,24 @@
+import { FileOutput, OutDir } from "@subsquid/util-internal-code-printer";
+import * as prettier from "prettier";
```
belopash, line 2: Let's prettify files by hands. I don't think it's that important to use prettier for that

evm/evm-typegen/package.json
```diff
@@ -24,12 +24,12 @@
     "@subsquid/util-internal": "^3.1.0",
     "@subsquid/util-internal-code-printer": "^1.2.2",
     "@subsquid/util-internal-commander": "^1.3.2",
-    "commander": "^11.1.0"
-  },
-  "peerDependencies": {
-    "ethers": "^6.9.0"
+    "@subsquid/evm-codec": "file:../evm-codec",
```
belopash, line 27: ref to local pacakge

PR #263, opened 2024-03-26
Use connection headers for for ws and http

Moliholy:
This PR adds the possibility to specify custom headers to be sent through both HTTP and WS connections. Useful in case the node requires some sort of authentication.


evm/evm-processor/src/processor.ts
```diff
@@ -36,6 +36,10 @@ export interface RpcEndpointSettings {
      * Maximum number of requests in a single batch call
      */
     maxBatchCallSize?: number
+    /**
+     * Connection headers
+     */
+    headers?: ConnectionHeaders
```
eldargab, line 42: Lets just use `Record<string, string>`

util/rpc-client/src/client.ts
```diff
@@ -38,7 +39,7 @@ export interface CallOptions<R=any> {
     validateError?: ErrorValidator<R>
 }
 
-
+export type ConnectionHeaders = {[key: string]: string}
```
eldargab, line 42: `Record<string, string>`

util/rpc-client/src/transport/http.ts
```diff
@@ -3,6 +3,7 @@ import {Logger} from '@subsquid/logger'
 import {fixUnsafeIntegers} from '@subsquid/util-internal-json-fix-unsafe-integers'
 import {RpcProtocolError} from '../errors'
 import {Connection, RpcRequest, RpcResponse} from '../interfaces'
+import {ConnectionHeaders} from '../client';
```
eldargab, line 6: No need for circular reference.

eldargab (changes requested):
Please, provide change file as well.

```
rush change -b origin/master
```

util/rpc-client/src/transport/http.ts
Moliholy: Removed.

util/rpc-client/src/client.ts
Moliholy: Fixed.

evm/evm-processor/src/processor.ts
Moliholy: Fixed.

Moliholy:
> Please, provide change file as well.

They are now included in the PR. Thanks for the tip.

PR #264, opened 2024-03-27
Make number functions accept both int and bigint, use strings instead of Buffer in bytes codec

vanruch:
Not exactly sure about this PR. I feel like working with strings is easier for squid users


belopash (commented):
Imho, for codec it's better to work with Uint8Arrays, but generated classes can accept both types and return hex strings

PR #266, opened 2024-04-08
map storage by pallet name instead of prefix

belopash:


eldargab (changes requested):
This is not correct as pallet prefix can be changed between runtimes?, yet it is hardcoded in the generated code

Here is a plan.

1 Change the storage description type to

```ts
interface Storage {
  [pallet: string]: {
     prefix: string
     items: {
       [name: string]: StorageItem
     }
  }
}
```

2 Update all `Runtime.getStorage()`, `Runtime.queryStorage()` and friends to expect `Pallet.Name` names and to use pallet prefix for key encoding.

3 Bump `@subsquid/substrate-runtime` to a new major version.
4 Adjust dependencies if needed.

belopash:
> This is not correct as pallet prefix can be changed between runtimes?, yet it is hardcoded in the generated code
> 
> Here is a plan.
> ....
Right, forgot about that.

By the way, which part should be backwards compatible? I thougth it was about `runtime.description.storage[prefix]`

substrate/substrate-runtime/src/metadata/runtime-description.ts
```diff
@@ -708,7 +721,14 @@ class FromOld {
                     docs: e.docs
                 }
             })
-            storage[mod.storage.prefix] = items
+            storage[mod.name] = {
+                prefix: mod.storage.prefix,
+                items
+            }
+
+            if (storage[mod.storage.prefix] == null) {
```
eldargab, line 729: This will generate extra items in "generate all" scenario and kinda pollutes metadata.

The same thing can be done within `Runtime` and typegen.

`private palletsByPrefix: {[prefix: string]: string}`

PR #274, opened 2024-05-03
Support both number and bignumber types when encoding numbers, add missing int/uint evm codecs

vanruch:


evm/evm-codec/src/codecs/primitives.ts
```diff
@@ -14,119 +17,119 @@ export const bool: Codec<boolean> = {
   isDynamic: false,
 }
 
-export const uint8: Codec<number> = {
-  encode(sink: Sink, val: number) {
-    sink.u8(val)
+export const uint8: Codec<Numberish, number> = {
+  encode(sink: Sink, val: Numberish) {
+    sink.u8(safeToNumber(val, 0, 255))
```
belopash, line 22: @eldargab can we just check here if it is safeInteger and sink (view) will de the rest?

belopash: just to reduce overhead

eldargab: Yes, `DataView` set methods already check input boundaries.

belopash: As turned out it doesn't

PR #276, opened 2024-05-09
Add address validation and checksum conversion in codec

vanruch:


evm/evm-codec/src/codecs/primitives.ts
```diff
@@ -210,7 +211,7 @@ export const address: Codec<string> = {
     sink.address(val)
   },
   decode(src: Src): string {
-    return src.address()
+    return toChecksumAddress(src.address())
```
belopash, line 214: Addresses in `tx.from`, `tx.to` and `log.address` are always in lower case. It will be more convinient to have decoded data in the same format to avoid uneccesary `.toLowerCase()` calls

vanruch: We kinda had it in this format when used ethers, so it brings back old behaviour. But let's go with what is more convenient

PR #304, opened 2024-06-27
evm: fetch multicall pages in parallel

belopash:


evm/evm-typegen/src/multicall.ts
```diff
@@ -34,133 +34,141 @@ export class Multicall extends ContractBase {
...
+      let splits = Math.ceil(left / maxSize)
+      let step = Math.round(left / splits)
+      yield [beg, beg + step]
+      beg += step
   }
 }
+
+function* splitArray<T>(maxSize: number, arr: T[]): Iterable<T[]> {
+  if (arr.length <= maxSize) {
+      arr
```
eldargab, line 168: `yield`

PR #305, opened 2024-07-03
AddFireblocksWeb3ProviderSupport

HelmiTrabelsi:
As we are paying for Fireblocks services, including the Fireblocks Web3 provider, we want to use this service in the indexer to avoid paying for another Web3 provider. However, we could not use it because squid-SDK only supports RPC URLs.

These changes allow the use of Fireblocks as a Web3 infrastructure provider in the Squid-SDK by making changes to the rpc-client and evm-processor packages. This update also facilitates using any other Web3 infrastructure provider that doesn't support RPC URLs.


PinkFromTheFuture (approved):
Code looks good, but the description of the PR could be improved

belopash:
This is not needed. Fireblocks SDK [uses](https://github.com/fireblocks/fireblocks-sdk-js/blob/30c39f8835f82423e2b332171ac622a664f221b6/src/api-client.ts#L15) axios with regular urls under the hood.

What you need to do is to pass [`baseUrl`](https://developers.fireblocks.com/docs/javascript-guide#your-first-fireblocks-javascript-script) to `url` [property](https://github.com/subsquid/squid-sdk/blob/669c2fc49a6745d53be6b11fac213634e30f9efd/evm/evm-processor/src/processor.ts#L22) and add ` X-API-Key` header with api key to the [`headers`](https://github.com/subsquid/squid-sdk/blob/669c2fc49a6745d53be6b11fac213634e30f9efd/evm/evm-processor/src/processor.ts#L42).

HelmiTrabelsi:
According to the Fireblocks SDK documentation (https://developers.fireblocks.com/reference/signing-a-request-jwt-structure) we need base url, X-API-Key, JWT. We tried to add this to our setRpcEndpoint but got 401 error. 

```
new EvmBatchProcessor().setRpcEndpoint({
    url: 'https://api.fireblocks.io/v1',
    rateLimit: 10,
    headers: {
      'X-API-Key': 'XXXXXXXXXXXXXXX',
      'Authorization': 'Bearer JWT'
  })
```

belopash:
@HelmiTrabelsi just to make sure, does fireblocks support rpc methods other than those releted to transactions signing? e.g eth_blockNumber

Initially I checked the wrong packag, but from the provider package you are importing, it looks like authorization for those methods are not needed at all, except some Authorization header formed from the creds in a url itself
https://github.com/fireblocks/fireblocks-web3-provider/blob/c0ee722f103548b26ef1317bf00f660d962cc603/src/provider.ts#L337
https://github.com/fireblocks/fireblocks-web3-provider/blob/c0ee722f103548b26ef1317bf00f660d962cc603/src/provider.ts#L75

PR #312, opened 2024-07-17
allow to load sibling calls

tmcgroul:


belopash:
@tmcgroul is it "all extrinsic calls" or "calls with the same parent call"?

tmcgroul:
@belopash calls with the same parent

belopash:
@tmcgroul maybe it is better to introduce "extrinsicCalls"? This could potentially cover more cases

substrate/substrate-processor/src/mapping.ts
```diff
@@ -219,6 +220,15 @@ export class Call {
         this.#subcalls = calls
     }
 
+    get siblings(): Call[] {
```
belopash, line 223: this new field looks a bit unnatural. You already can achieve the same by using `call.parent.subcalls`

tmcgroul:
closing due to "some" open questions and revealing that the task isn't really necessary 🤷‍♂️

PR #315, opened 2024-07-22
add finalityConfirmation to substrate processor

tmcgroul:


substrate/substrate-data-raw/src/datasource.ts
```diff
@@ -42,20 +42,50 @@ export class RpcDataSource {
...
+        if (this.finalityConfirmation == null) {
+            let head = await this.rpc.getFinalizedHead()
+            let header = await this.rpc.getBlockHeader(head)
+            assert(header, 'finalized blocks must be always available')
+            return qty2Int(header.number)
+        } else {
+            let retries = 0
+            while (retries < 5) {
+                let head = await this.rpc.getHead()
+                let header = await this.rpc.getBlockHeader(head)
```
belopash, line 66: `chain_getHeader` can work without hash, so there is probably no need to call the two methods sequentially

tmcgroul: yes, that's right. thanks for the suggestion!

substrate/substrate-data-raw/src/datasource.ts
```diff
@@ -42,20 +42,41 @@ export class RpcDataSource {
...
+    }
+
+    async getFinalizedHead(best: string): Promise<string> {
+        if (this.finalityConfirmation == null) {
+            return this.rpc.getFinalizedHead()
+        } else {
+            let header = await this.rpc.getBlockHeader(best)
+            assert(header)
+            let height = qty2Int(header.number) - this.finalityConfirmation
+            let hash = await this.rpc.getBlockHash(height)
```
belopash, line 76: `best` is already blockHash, so you basically do `hash` -> `height` -> `hash`

substrate/substrate-data-raw/src/datasource.ts
```diff
@@ -42,20 +42,41 @@ export class RpcDataSource {
...
+            assert(header, 'finalized blocks must be always available')
+            return qty2Int(header.number)
+        } else {
+            let header = await this.rpc.getBlockHeader()
+            assert(header, 'the header of the latest block on the chain must be always available')
+            return Math.max(0, qty2Int(header.number) - this.finalityConfirmation)
+        }
+    }
+
+    async getFinalizedHead(best: string): Promise<string> {
```
belopash, line 69: should `best` be optional?

substrate/substrate-data-raw/src/datasource.ts
```diff
@@ -42,20 +42,41 @@ export class RpcDataSource {
...
+        if (this.finalityConfirmation == null) {
+            return this.rpc.getFinalizedHead()
+        } else {
+            let header = await this.rpc.getBlockHeader(best)
+            assert(header)
+            let height = qty2Int(header.number) - this.finalityConfirmation
+            let hash = await this.rpc.getBlockHash(height)
+            assert(hash)
```
belopash, line 77: retry is needed

PR #317, opened 2024-07-22
feat: add s3 ops metrics

octo-gone:


util/util-internal-dump-cli/src/prometheus.ts
```diff
@@ -56,7 +58,21 @@ export class PrometheusServer {
                     kind: 'failure'
                 }, metrics.connectionErrors)
             }
-        });
+        })
+
+        this.s3RequestsGauge = new Gauge({
```
mo4islona, line 63: It should be counter

octo-gone: Probably Guage suits here better. We call `set` instead of `inc` because we get exact value of metric from `S3Fs.getMetrics`. Incremet will require to calculate increase or change S3Fs significantly.

mo4islona: We won't be able to count requests normally on restarts with gauge, which is always going to start from 0

mo4islona (approved):

PR #321, opened 2024-07-23
fix subcalls relation

tmcgroul:


substrate/substrate-processor/src/mapping.ts
```diff
@@ -405,6 +405,10 @@ export function setUpItems(block: Block): void {
                 rec.parentCall = prev
                 populateSubcalls(prev, rec)
             }
+            if (prev.parentCall && isSubcall(prev.parentCall, rec)) {
```
belopash, line 408: nested calls are handled incorrectly.

[1, 1, 1],  [1, 1], [1]  -> the first will be child of the second, the second will be cild of the third

[1, 1, 1],  [1]  -> the first will be child of the third

tmcgroul: yeah, i also expect that it works like that

tmcgroul: but i don't see the problem...

tmcgroul: ok if we talk about `[1, 1, 1], [1]` then there will be wrong parent assignment...
the parent will have the child as a subcall but i believe `[1]` shouldn't be in `.parentCall`

belopash: idk that should be the correct behavior here, but `isSubcall` assumes that `[1, 1, 1]` is a subcall of `[1]`

tmcgroul: it is expected according to subcalls query - https://github.com/subsquid/archive.py/blob/master/sqa/substrate/query.py#L422-L426

substrate/substrate-processor/src/mapping.ts
```diff
@@ -487,3 +493,8 @@ function isSubcall(parent: CallKey, call: CallKey): boolean {
     }
     return true
 }
+
+
+function traverseCalls(prev: Call, call: Call) {
```
belopash, line 498: ?

substrate/substrate-processor/src/mapping.ts
belopash: > it is expected according to subcalls query -
Bot sure, I guess it only says "request all subcalls and thier child", but nothing about how they should be grouped

substrate/substrate-processor/src/mapping.ts
tmcgroul: somebody did it. not me 💅

substrate/substrate-processor/src/mapping.ts
tmcgroul: yeah, another way is to put only direct "children" into `subcalls` array. i guess both options are fine...
@eldargab how subcalls are supposed to work?

tmcgroul: i got confirmation that subcalls represent all children

PR #333, opened 2024-08-12
Added starknet support, Subsquid Network Gateway ingest

Wizard1209:
Added packages for starknet index, and test indexer for transactions


tmcgroul (changes requested):

starknet/starknet-normalization/src/data.ts
```diff
@@ -0,0 +1,49 @@
+type FELT = string
+
+export interface BlockHeader {
+    hash: FELT
```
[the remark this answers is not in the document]
Wizard1209: FELT is the default byte array type in starknet.
In starknet-data, I have copied the structure from the specification, which highlights hashes, but still implies FELT like this
https://github.com/starkware-libs/starknet-specs/blob/b5c43955b1868b8e19af6d1736178e02ec84e678/api/starknet_api_openrpc.json#L1317
In starknet-normalization I simplified it because it doesn't need to be that specific imo.

starknet/starknet-stream/src/archive/data-schema.ts
```diff
@@ -0,0 +1,99 @@
...
+    taggedUnion,
+    oneOf,
+    constant
+} from '@subsquid/util-internal-validation'
+import {FieldSelection} from '../data/model'
+import {Selector} from '../data/util'
+
+
+
+// TODO: should i add separate FELT validator
```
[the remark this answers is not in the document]
Wizard1209: FELT is limited in size, but I'm not sure if validating it makes any sense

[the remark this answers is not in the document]
tmcgroul: we don't do in other cases so here we can skip it as well =)

starknet/starknet-stream/src/data/model.ts
```diff
@@ -0,0 +1,61 @@
...
+import type {
+    BlockRequiredFields,
+    EventRequiredFields,
+    TransactionRequiredFields
+} from './data-partial'
+
+export type FELT = string
+
+
+export interface FieldSelection {
```
tmcgroul, line 12: currenlty FieldSelection allows to specify fields that aren't available in the gateway. i tried to request all fields and got this error message:
{"body":{"title":"400 Bad Request","description":"{'fields': {'block': defaultdict(<class 'dict'>, {'l1GasPriceInFri': {'key': ['Invalid value.']}, 'l1GasPriceInWei': {'key': ['Invalid value.']}, 'starknetVersion': {'key': ['Invalid value.']}}), 'transaction': defaultdict(<class 'dict'>, {'blockNumber': {'key': ['Invalid value.']}}), 'event': defaultdict(<class 'dict'>, {'blockNumber': {'key': ['Invalid value.']}, 'key0': {'key': ['Invalid value.']}, 'key1': {'key': ['Invalid value.']}, 'key2': {'key': ['Invalid value.']}, 'key3': {'key': ['Invalid value.']}, 'keysSize': {'key': ['Invalid value.']}, 'restKeys': {'key': ['Invalid value.']}})}}"}}

field selection should match to the one that is defined in archive.py
https://github.com/subsquid/archive.py/blob/master/sqa/starknet/query.py#L43

tmcgroul (changes requested):

starknet/starknet-stream/src/data/model.ts
Wizard1209: Thanks a lot, I have initially built data models according to tables in archive.py and havent rechecked since. For events I will need one more workaround because required fields fundamentally different from actual structure. For current implementation this means optional keys will always be presented in object received in handler.
https://github.com/Wizard1209/squid-sdk/blob/starknet-support/starknet/starknet-stream/src/data/model.ts#L52

tmcgroul: i guess we can keep keys as an array. squid developers will have to deal with indexes on their own.
my proposal is to keep EventRequest as it is. but starknet-normalization shouldn't split keys into separate fields and Event entity should contain only keys field.
that's what we do with log topics in evm

starknet/starknet-stream/src/archive/data-schema.ts
```diff
@@ -0,0 +1,94 @@
...
+    let Transaction = object({
+        transactionIndex: NAT,
+        ...project(fields.transaction, {
+            blockNumber: option(NAT),
+            transactionHash: option(FELT),
+            contractAddress: option(FELT),
+            entryPointSelector: option(FELT),
+            calldata: option(array(FELT)),
+            maxFee: option(FELT),
+            type: oneOf({
```
tmcgroul, line 43: should be extended with `L1_HANDLER`
```
18:08:39 FATAL sqd:batch-processor DataValidationError: invalid value at /779/transactions/1/type: given value does not match any of the expected patterns:
                                       INVOKE: value "L1_HANDLER" is not equal to "INVOKE"
                                       DECLARE: value "L1_HANDLER" is not equal to "DECLARE"
                                       DEPLOY_ACCOUNT: value "L1_HANDLER" is not equal to "DEPLOY_ACCOUNT"
                                       DEPLOY: value "L1_HANDLER" is not equal to "DEPLOY"
                                       at cast (/Users/sar/projects/squid-sdk/util/util-internal-validation/lib/dsl.js:80:15)
                                       at StarknetGateway.getBlockStream (/Users/sar/projects/squid-sdk/starknet/starknet-stream/lib/archive/source.js:45:62)
                                       at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
                                       at async StarknetDataSource.getBlockStream (/Users/sar/projects/squid-sdk/starknet/starknet-stream/lib/source.js:183:32)
                                       at async Processor.run (/Users/sar/projects/squid-sdk/processor/batch-processor/lib/run.js:72:24)
```

starknet/starknet-normalization/src/data.ts
```diff
@@ -0,0 +1,39 @@
...
+    height: number
+    parentHash: FELT
+    status: string
+    newRoot: FELT
+    timestamp: number
+    sequencerAddress: FELT
+}
+
+export type TransactionType = 'INVOKE' | 'DECLARE' | 'DEPLOY_ACCOUNT' | 'DEPLOY'
```
tmcgroul, line 13: L1_HANDLER

tmcgroul (changes requested):

starknet/starknet-objects/src/items.ts
```diff
@@ -0,0 +1,118 @@
...
+    eventIndex: number
+    transactionIndex: number
+    #block: BlockHeader
+    #transaction?: Transaction
+
+    constructor(
+        block: BlockHeader,
+        e: PartialEvent
+    ) {
+        this.id = formatId(block, e.eventIndex)
```
tmcgroul, line 92: one last thing:
id should be calculated like this `formatId(block, e.transactionIndex, e.eventIndex)`. otherwise events will not have unique id

tmcgroul (changes requested):

starknet/starknet-objects/src/items.ts
Wizard1209: thanks, I shouldn't have missed this.

Issue #334, opened 2024-08-13
Substrate `storage.<pallet>.<item>.<version>.is(block)` false positive


khssnv:
**Describe the bug**

The call `storage.mypallet.myitem.v1.is(blockHeader)` (where `storage` is imported from `types` generated by `squid-substrate-typegen`) returns `true` at block at height where it expected to return `true` for `v2` (which added new fields to `myitem` type) only.

**To Reproduce**

1. Setup the project.

    ```console
    git clone --branch debug/khssnv-squid-sdk-bug-report https://github.com/Cerebellum-Network/cere-squid-indexer.git
    cd cere-squid-indexer
    npm i
    ```

1. Launch processor.

    ```console
    docker compose up -d db
    sqd process
    ```

1. See error message.

    ```console
    ...
    Unexpected storage layout version selection. Selected v48013 for spec version 54116 which is higher than the next v48017.
    ...
    ```

Please tell if this one is too large and I'll prepare a minimal reproducible example.

**Expected behavior**

In my specific case with the project set up above, the [storage.ddcCustomers.buckets.v48013.is(block)](https://github.com/Cerebellum-Network/cere-squid-indexer/blob/55e6dd9da9f60ecbcb6d8ab498e7e264befbea51/src/processors/ddcBucketsProcessor.ts#L30) call is expected to return `false` instead of `true` at height `5,238,731` and return `true` at [storage.ddcCustomers.buckets.v54100.is(block)](https://github.com/Cerebellum-Network/cere-squid-indexer/blob/55e6dd9da9f60ecbcb6d8ab498e7e264befbea51/src/processors/ddcBucketsProcessor.ts#L87C20-L87C65).

**Environment (please complete the following information):**
 - specify the version of the core packages: `"@subsquid/substrate-processor": "8.3.0"`, `"@subsquid/substrate-typegen": "8.1.0"`, `"@subsquid/substrate-metadata-explorer": "3.2.0"`
 - node.js version `16`
 - npm version `10.8`
 - OS version `Linux x1 6.5.0-1027-oem #28-Ubuntu SMP PREEMPT_DYNAMIC Thu Jul 25 13:32:46 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux`
 - Reproducible example or a repo link https://github.com/Cerebellum-Network/cere-squid-indexer

**Applicable to decoding issues:**

 - chain name `cere-dev`
[...]

PR #337, opened 2024-08-19
substrate: fix structs comparison false positive

khssnv:
Fixes https://github.com/subsquid/squid-sdk/issues/334.


eldargab:
This patch will break call parsing procedures.

@belopash has a better plan:  leave it as is for the "latest" version and make it strict for the older versions.

PR #347, opened 2024-10-11
feat(typeorm-store): add optionnal head and rollback db hooks for typeorm store

0x666c6f:
The goal of this PR is to enable optional hooks on the `typeorm` store to allow external services relying on subsquid to get notified when a new head, or a rollback happened.


dzhelezov:
Closed as it is more appropriate to handle this logic directly from the processor w/o changing the interface

PR #349, opened 2024-10-29
solana: add instructionIndex to instruction

tmcgroul:


tmcgroul:
@eldargab should `innerInstructionIndex` be `undefined` for primary instructions or `-1`?

eldargab:
I think, it should be undefined.

solana/solana-normalization/src/mapping.ts
```diff
@@ -396,7 +397,14 @@ class InstructionParser {
...
         }
 
+        if (stackHeight === 1) {
+            this.innerPos = -1
+        } else {
+            this.innerPos++
+        }
+
         let i: Instruction = {
+            innerInstructionIndex: this.innerPos,
```
eldargab, line 407: I think, it is better to set it directly, rather than recalculating it here.

eldargab (changes requested):

solana/solana-normalization/src/mapping.ts
tmcgroul: do you mean outside of `push` method? it seems that inner instructions can be created in several methods so it's hard to track them 🤷‍♂️

PR #355, opened 2024-11-16
evm: add tx.authorizationList field

tmcgroul:


evm/evm-processor/src/interfaces/evm.ts
```diff
@@ -35,6 +35,16 @@ export interface EvmTransaction extends _EvmTx, _EvmTxReceipt {
 }
 
 
+export interface Authorization {
```
tmcgroul, line 38: @eldargab do we need `Evm` prefix as for other entities?

evm/evm-processor/src/mapping/schema.ts
```diff
@@ -64,6 +74,7 @@ export function getTxProps(fields: FieldSelection['transaction'], forArchive: bo
             s: withSentinel('Transaction.s', '0x', BYTES),
             yParity: option(natural),
             chainId: option(natural),
+            authorizationList: option(array(Authorization)),
```
tmcgroul, line 77: @eldargab do we need ability to select each field manually or it's fine and we can load the whole array of structures?

eldargab: Let's expose it as a whole.

evm/evm-processor/src/interfaces/evm.ts
eldargab: Prefix might be helpful, but don't forget to re-export `Authorization` in a prefix-less form.

I'm afraid `Authorization` is too ambiguous, though. May be call it like `EIP7702Authorization`?

PR #372, opened 2025-02-17
Solana data service

eldargab:
This PR has the following issues:

* [Transaction errors are not decoded, when they come from Geyser plugin](https://github.com/subsquid/squid-sdk/blob/d7c73853c6da78366bf405247ea4c9cd3c04f4d0/solana/solana-data-service/src/data-source/geyser.ts#L249)
* [Validate RewardType strings against native RPC](https://github.com/subsquid/squid-sdk/blob/d7c73853c6da78366bf405247ea4c9cd3c04f4d0/solana/solana-data-service/src/data-source/geyser.ts#L303)

I don't intend to fix the above myself, but will promote this PR from `Draft` status once it will be ready otherwise.


eldargab:
@tmcgroul 

`solana-dump` and `solana-ingest` are ready to produce a new slot-based Solana archive.

Other dumpers and ingesters should not be broken by this PR, but I haven't checked that!

@belopash 

I have suspicion, that `Error` (enumerable) properties are not properly transferred across message ports. Could you check it?

PR #389, opened 2025-04-18
Feat block metrics

DamianFigiel:


eldargab:
It is not dumper's job to monitor RPC state.

PR #405, opened 2025-07-04
solana: backport new block normalization

belopash:


solana/solana-stream/src/rpc/mapping.ts
```diff
@@ -4,10 +4,19 @@ import {PartialBlock} from '../data/partial'
...
+        throw addErrorContext(new Error(msg), props)
+    },
+    error: function(props: any, msg: string): void {
+        throw addErrorContext(new Error(msg), props)
+    }
+}
 
 export function mapBlock(src: rpc.Block, req: DataRequest): PartialBlock {
-    let block = mapRpcBlock(src)
+    let block = mapRpcBlock(src, failingJournal)
```
eldargab, line 19: We shall use "logging journal" to not fail on log parsing errors.

eldargab (changes requested):

PR #433, opened 2026-02-11
Add evm-ingest option to fix log index DATA-85

kalabukdima:
Stable chain has wrong log index numbering, it goes
`[0, 1, 2, 3], [4, 1]`

Adding a workaround which is disabled by default

### DIFF
--- a/common/config/rush/pnpm-lock.yaml
+++ b/common/config/rush/pnpm-lock.yaml
@@ -3516,6 +3516,12 @@ packages:
       undici-types: 6.21.0
     dev: false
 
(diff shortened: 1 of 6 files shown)


kalabukdima:
> or maybe even better we won't change raw data and will keep it as we get it from rpc we just fix indexes at normalization step

Agree, it's always worth storing raw RPC responses unmodified. I intended to only affect the ingester and data service, but maybe added it in an extra location

tmcgroul:
closing as this pr targets open-beta which is no longer used

PR #461, opened 2026-04-02
http-client: replace `node-fetch` with native `fetch` API

belopash:


util/http-client/src/client.ts
```diff
@@ -499,3 +507,15 @@ export function asRetryAfterPause(res: HttpResponse | Error): number | undefined
 }
 
 const HTTP_DATE_REGEX = /^(?:Mon|Tue|Wed|Thu|Fri|Sat|Sun), (\d{2}) (Jan|Feb|Mar|Apr|May|Jun|Jul|Aug|Sep|Oct|Nov|Dec) (\d{4}) (\d{2}):(\d{2}):(\d{2}) GMT$/;
+
+const ACCEPT_ENCODING = detectAcceptEncoding()
+
+function detectAcceptEncoding(): string | undefined {
```
mo4islona, line 513: Have you tested this in the Node version that supports ZSTD but doesn't parse response ZSTD body automatically?

PR #467, opened 2026-04-24
evm: topicSelection helper, simplified factories, modular codegen, faster encode/decode

belopash:
## `@subsquid/evm-abi`

- **New `AbiEvent.topicSelection(filter?)`** for building `LogRequestWhere`-compatible topic filters. `topic0` is always set to the event's topic; for each indexed field present in `filter`, the provided values are ABI-encoded to 32-byte words and placed at the corresponding `topic1`/`topic2`/`topic3` position. Multiple values per field are OR-ed by the gateway.
  ```ts
  // all Transfers from a contract
  processor.addLog({ where: { address: [CONTRACT], ...Transfer.topicSelection() } })

  // Transfers to specific recipients
  processor.addLog({ where: { address: [CONTRACT], ...Transfer.topicSelection({ to: [ADDR_A, ADDR_B] }) } })
  ```
  New exported types: `TopicFilter`, `IndexedTopicFilter`, `EventArgumentsInput`.

- **Simplified factory signatures.** The now-unused `signature` positional argument has been dropped:
  - `event(topic, signature, args)` → `event(topic, args)`
  - `fun(selector, signature, args, returnType?)` → `fun(selector, args, returnType?)`
  - `viewFun(...)` analogously.

  `fun` / `viewFun` also accept a `Codec` (including `struct({...})`) as the return type rather than a bare struct literal — wrap multi-output return shapes with `p.struct({...})`.

- Hand-written ABI bindings against the previous API need to be updated; bindings produced by `evm-typegen` are regenerated automatically (see below).

## `@subsquid/evm-typegen`

- **Modular per-contract output.** Instead of a single flat `myContract.ts`, the generator now emits a folder:
  ```
  abi/myContract/
    contract.ts    // Contract class
    events.ts      // one `export const <Name>` + type per event
    functions.ts   // one `export const <name>` + Params/Return types per function
    index.ts       // re-exports: `Contract`, `events.*`, `functions.*`
  ```
  The `index.ts` re-exports `events` and `functions` as namespaces, so existing imports like `import { events, functions, Contract } from './abi/myContract'` keep working. Individual exports (`import { Transfer } from './abi/myContract/events'`) are now also possible and tree-shake cleanly.

- **Per-export JSDoc.** Each generated event/function is prefixed with its canonical Solidity signature as a comment (e.g. `// Transfer(address,address,uint256)`), which previously lived as a positional string argument.

## Benchmarks

Added `test/evm-codec-bench` comparing the new implementation against the previously published `@subsquid/evm-codec`/`@subsquid/evm-abi` (`0.3.x`) and `viem`'s ABI helpers. Parity is asserted before timing.

`AbiEvent.decode` — ERC-20 `Transfer` (200k iters, Node v20):

| path | ops/s | vs old | vs viem |
|---|---:|---:|---:|
| new (JIT) | 6,314,649 | 15.32× | 36.58× |
| old (loop) | 412,296 | — | — |
| viem | 172,628 | — | — |

`AbiFunction.encode/decode` — dynamic 4-arg call (50k iters):

| op | new | old | viem | new vs old | new vs viem |
|---|---:|---:|---:|---:|---:|
| encode | 187,192 | 112,895 | 30,554 | 1.66× | 6.13× |
| decode | 358,199 | 104,783 | 20,622 | 3.42× | 17.37× |

Struct-level encode/decode (a representative sample):

| shape | op | new vs old | new vs viem |
|---|---|---:|---:|
| static (`Transfer` args) | encode / decode | 1.70× / 7.14× | 1.26× / 25.68× |
| dynamic (bytes + array)  | encode / decode | 2.12× / 1.93× | 3.19× / 10.69× |
| nested tuple (Uniswap swap) | encode / decode | 2.55× / 4.45× | 4.62× / 13.02× |
| aggregate-like (multicall tryAggregate output) | encode / decode | 1.19× / 4.83× | 3.07× / 17.06× |

Full numbers and harness: [`test/evm-codec-bench/README.md`](test/evm-codec-bench/README.md).


mo4islona:
## Review: Bugs & Correctness Issues

Nine issues found across three model reviews (Opus, Sonnet, GPT-5.4). Four are blocking.

---

### 🔴 BLOCKER 1 — `topicSelection()` silently truncates multi-slot static indexed types

**File:** `evm/evm-abi/src/abi-components/event.ts` — `topicSelection()`

```ts
// topicSelection() — wrong:
const sink = new HexSink(1)           // hardcoded 1

// createEncodeInline() — correct, already uses slotsCount:
const slots = codec.slotsCount ?? 1
```

`HexSink.toString()` returns exactly `slots × 32` bytes. For static indexed types with `slotsCount > 1` (e.g. `indexed(fixedSizeArray(uint256, 3))`), `topicSelection()` silently returns the first 32 bytes only, while `encode()` writes the full 96 bytes. The filter will never match any real log.

**Fix:** `new HexSink(codec.slotsCount ?? 1)` — same as `createEncodeInline` already does.

There's also no test cross-checking `topicSelection({field: [v]}).topicN[0]` against `encode({field: v}).topics[N]`, which would have caught this.

---

### 🔴 BLOCKER 2 — Static indexed tuples/fixed arrays are not keccak256-hashed (ABI spec violation)

**File:** `evm/evm-abi/src/abi-components/event.ts` — constructor

```ts
const param = arg.indexed && arg.isDynamic
    ? ({...bytes32, isDynamic: true, indexed: true} as any)
    : arg   // ← static structs and fixed arrays fall through untouched
```

Per the [Ethereum ABI spec](https://docs.soliditylang.org/en/latest/abi-spec.html#indexed-event-encoding), **all indexed arrays and structs — even statically-sized ones** — must be keccak256-hashed into a single 32-byte topic. The `isDynamic` check only catches dynamically-sized types (string, bytes, dynamic arrays). A codec like `indexed(struct({a: uint256, b: uint256}))` is not `isDynamic` but per spec its topic must be `keccak256(abi_encode(a, b))`, not the raw 64 bytes.

**Impact:** produces non-spec topics in both `encode()` and `topicSelection()`, making logs undecodable and filters non-functional for this combination.

**Fix:** the gate should also cover `struct` and `fixedSizeArray` codecs — check `slotsCount > 1` or add explicit type-kind flags.

---

### 🔴 BLOCKER 3 — `myContract.ts` silently shadows `myContract/index.ts` on upgrade

**File:** `evm/evm-typegen/src/main.ts`

TypeScript resolves `import from './abi/myContract'` as `myContract.ts` **before** `myContract/index.ts`. Without `--clean`, users who upgrade and re-run typegen will have both files on disk. TypeScript silently binds to the stale old file — no error, no warning, builds pass.

`--clean` deletes the entire output directory (too destructive if users colocate other files there).

**Fix:** for each `spec.name`, detect and delete the legacy `${basename}.ts` with a logged warning before writing the new `${basename}/` directory.

---

### 🔴 BLOCKER 4 — `Contract` wraps non-view functions that have outputs (`stateMutability` regression)

**File:** `evm/evm-typegen/src/typegen.ts` — `generateContract()` / `description.ts`

```ts
const viewFns = this.contract.functions.filter((f) => f.outputs.length > 0)
```

`FunctionDef` doesn't carry `stateMutability`. So `transfer(address,uint256) returns (bool)` passes the filter and gets a `Contract.transfer()` method calling `eth_call`. This silently simulates a state-changing call instead of sending a transaction — a regression from the old API where `viewFun` vs `fun` made the distinction explicit.

**Fix:** carry `stateMutability` through `FunctionDef` in `description.ts` and filter by `'view' | 'pure'` in `generateContract()`.

---

### 🟡 HIGH — `topicSelection({field: []})` produces a match-nothing filter silently

**File:** `evm/evm-abi/src/abi-components/event.ts` — `topicSelection()`

```ts
result[keys[i]] = values.map(...)   // values = [] → topic1: []
```

In Ethereum log filter semantics, `topicN: []` means "OR over zero values" = **matches no logs** — not "ignore this topic position". A caller who conditionally builds a filter and passes an empty array gets zero results with no error.

**Fix:** `if (values.length === 0) continue` (treat same as omitted/null).

---

### 🟡 HIGH — `__proto__` field name corrupts JIT-decoded objects

**File:** `evm/evm-abi/src/abi-components/event.ts` + `function.ts` — `createDecodeInline()`

```ts
fields.push(`${propName(o.key)}:__v${n}`)
```

If `propName('__proto__')` returns the bare identifier, the generated function body contains `return {__proto__: __v0, ...}`, which sets the object prototype rather than a data property. `__proto__` is a valid Solidity identifier so this is reachable with real ABIs.

**Fix:** ensure `propName`/`propAccess` emit computed string keys for `__proto__` (and `constructor`, `valueOf` etc. if not already handled).

---

### 🟠 MEDIUM — `TopicFilter` / `IndexedTopicFilter` not re-exported from generated `abi.support.ts`

**File:** `evm/evm-typegen/src/abi.support.ts`

```ts
export type {EventParams, EventArgumentsInput, FunctionArguments, FunctionReturn} from '@subsquid/evm-abi'
// TopicFilter, IndexedTopicFilter — absent
```

`abi.support.ts` is the sole import target for generated code. Users who want to type-annotate a `topicSelection()` result or write a helper accepting `IndexedTopicFilter<typeof MyEvent>` must reach into `@subsquid/evm-abi` directly.

**Fix:**
```ts
export type {EventParams, EventArgumentsInput, FunctionArguments, FunctionReturn,
             TopicFilter, IndexedTopicFilter} from '@subsquid/evm-abi'
```

---

### 🟠 MEDIUM — `EventDecodingError` reports `rec.data` even when a topic field fails

**File:** `evm/evm-abi/src/abi-components/event.ts` — `createDecodeInline()`

```ts
// catch always uses rec.data:
'}catch(e){throw new EventDecodingError(TOPIC,FIELD_NAMES[__i],rec.data,e.message);}'
```

When decoding fails on a topic field (decoded from `rec.topics[i+1]`), the error prints `rec.data` — the completely unrelated non-indexed payload. Makes debugging topic corruption impossible.

**Fix:** build a parallel `SOURCES` array bound alongside `FIELD_NAMES`, containing `rec.data` for data fields and `rec.topics[k]` for topic fields. Use `SOURCES[__i]` in the catch.

---

### 🟢 LOW — Empty `Contract` class always generated and exported for events-only ABIs

**File:** `evm/evm-typegen/src/typegen.ts`

`generateContract()` and the `export { Contract }` line in `writeIndex()` run unconditionally. For event-only ABIs (no functions), this emits a useless empty class. Should be gated behind `viewFunctions.length > 0` (post the `stateMutability` fix).

PR #488, opened 2026-05-26
Fix/http client six retry attempts by default

abernatskiy:


util/http-client/src/client.ts
```diff
@@ -75,7 +75,7 @@ export class HttpClient {
         this.setBaseUrl(options.baseUrl)
         this.agent = options.agent || defaultAgentProvider
         this.retrySchedule = options.retrySchedule || [10, 100, 500, 2000, 10000, 20000]
-        this.retryAttempts = options.retryAttempts || 0
+        this.retryAttempts = options.retryAttempts || 6
```
eldargab: As far as I remember some places assume absence of retries.

belopash: It doesn't sound correct that some components rely on the implicit number of attempts (but I don't remember such places). If so, they should explicitly set it then. Or you didn't mean default value?

belopash: should be `??` not `||`

abernatskiy:
Now the default is applied on the portal client level instead

PR #494, opened 2026-06-16
evm-rpc: include consensusContext in Tempo block hash

elina-chertova:
## Problem

The `evm-data-service` pods for **tempo-mainnet** hotblocks (both `uniblock` and `dwellir` providers) are in `CrashLoopBackOff`, each crashing on:

```
AssertionError [ERR_ASSERTION]: failed to verify block hash
    at Rpc.mapBlock (evm/evm-rpc/lib/rpc.js:119)
```

Because the metric `sqd_hotblocks_last_block{job="evm-*-tempo-mainnet-hotblocks-service"}` is therefore never emitted, the `absent(...)` hotblocks alerts (`No_Block_Data`) fire for every Tempo provider.

## Root cause

Tempo extended its block header (`TempoHeader`) with a **trailing optional `consensus_context`** field that is present on **post-fork** blocks:

```
rlp([general_gas_limit, shared_gas_limit, timestamp_millis_part, rlp([...ethereum_header_fields]), consensus_context?])
```

where `consensus_context = [epoch, view, parent_view, proposer]`
(see https://github.com/tempoxyz/tempo/blob/main/crates/primitives/src/header.rs — `#[rlp(trailing)]`).

`tempoBlockHash` only encoded the first four elements, so for any block that carries a `consensusContext` the computed hash differs from the chain's hash and `verifyBlockHash` throws. Pre-fork blocks (no `consensusContext`) hashed correctly, which is why the existing fixture `tempo-mainnet/17708868` passes and the regression went unnoticed.

## Fix

- Add `consensusContext` to the `GetBlock` validation schema — without it the field is stripped before reaching `tempoBlockHash`.
- Append the `consensus_context` RLP list (`[epoch, view, parentView, proposer]`) to `tempoBlockHash` when present. Trailing/optional: pre-fork blocks are unaffected.
- Add a regression test + post-fork `tempo-mainnet/25403305` fixture.

## Testing / evidence

Computed `tempoBlockHash` was checked against the live chain hash returned by RPC:

| block | has consensusContext | result |
|---|---|---|
| 17708868 (pre-fork fixture) | no | ✅ match (unchanged) |
| 25403018 (dwellir crash block) | yes | ✅ match |
| 25403237 | yes | ✅ match |
| 25403305 (new fixture) | yes | ✅ match |
| 25404009 (head) | yes | ✅ match |

Before the fix, all post-fork blocks mismatched (reproducing the pod crash); after, all match.

## Falsification

If a post-fork Tempo block's hash still fails verification after this change, the assumption that the header is exactly `[gas, gas, ts, inner, [epoch, view, parentView, proposer]]` is wrong (e.g. Tempo added another trailing header field). Re-check `tempoxyz/tempo` `crates/primitives/src/header.rs`.

## Follow-up (ops)

The fix only takes effect once a new `evm-data-service` image built from this change is rolled out to the Tempo hotblocks pods. This restores both confirmed cluster members (uniblock + dwellir tempo-mainnet).


tmcgroul:
already fixed here - https://github.com/subsquid/squid-sdk/commit/e56f20a917e727b98ac9a2e29a428286ddb1fa43

PR #498, opened 2026-06-17
Improve efficiency of array manipulation in multicall.ts in eth_call promises

aspnxdd:
## Summary

Optimizes the generated EVM multicall helper by avoiding extra array materialization and `flat()` when processing paged calls.

I've also added a JS file to bench it, the results are of a ~100% improvement (2x) in speed.


aspnxdd:
@abernatskiy hey bud, just tagging you here for visibility

PR #536, opened 2026-07-15
feat(evm/fallback): standalone-capable Portal/RPC source builders

abernatskiy:
## What

Makes `EvmPortalDataSourceBuilder` and `EvmRpcDataSourceBuilder` **full standalone data-source builders** — the familiar `DataSourceBuilder` shape (`setFields`/`addLog`/… + `.build()`) — while keeping the fallback's "define the query once" guarantee.

Previously they were minimal connection-only configs: they couldn't build a source on their own, and weren't reusable outside a fallback. Now the same builder works both ways:

```ts
// Standalone — returns an EVMDataSource<F>, exactly like DataSourceBuilder.build()
const rpcSource = new EvmRpcDataSourceBuilder()
    .setRpc({url: RPC_URL, network: 'ethereum-mainnet'})
    .setFields({log: {topics: true, data: true}})
    .addLog({where: {address: [CONTRACT], topic0: [TRANSFER]}, range: {from: 10_000_000}})
    .build()

// In a fallback — connection config only; the fallback injects the shared query
const fb = new EvmFallbackDataSourceBuilder()
    .setDownstreamSources([
        new EvmPortalDataSourceBuilder().setPortal(PORTAL_URL),
        new EvmRpcDataSourceBuilder().setRpc({url: RPC_URL, network: 'ethereum-mainnet'}),
    ])
    .setFields({log: {topics: true, data: true}})
    .addLog({where: {address: [CONTRACT], topic0: [TRANSFER]}, range: {from: 10_000_000}})
    .build()
```

## Design: "standalone builders, query injected"

The core fallback invariant is that every source fetches **identical** data. This PR keeps that but relocates the guard from compile-time to init-time (a deliberate, chosen trade-off):

- The shared query still lives once on `EvmFallbackDataSourceBuilder` and is injected into each downstream via `buildSource(fields, requests)`.
- A source used in a fallback that **also** sets its own field selection/query is rejected at `build()` time (`hasOwnQuery` guard) — the query must live in exactly one place.

## Structure

- Extract the shared query surface (`setFields`/`addLog`/`addTransaction`/`addTrace`/`addStateDiff`/`addQuery`/`includeAllBlocks`/`setBlockRange` + `getRequests`) into an exported abstract `EvmQueryBuilder` base, reused by all three builders — no duplication.
- Portal builds standalone via the canonical `DataSourceBuilder` (byte-identical output); RPC via `evmRpcStream` (lazy peer load unchanged).

## Testing

- 84 fallback tests pass (added: standalone Portal build, standalone + injected RPC build, conflict-throws).
- The RPC-build unit tests stub the lazy loader — `require('../rpc')` resolves a directory to `lib/index.js` only when compiled, not to the `.ts` source under vitest; the network-gated e2e covers the real RPC stack.
- `tsc --noEmit` + declaration emit clean; type inference verified end-to-end (`block.logs[0].topics: string[]`, `block.transactions[0].gasUsed: bigint`).

### DIFF
--- a/common/changes/@subsquid/squid-sdk/feat-standalone-fallback-source-builders_2026-07-15-10-44.json
+++ b/common/changes/@subsquid/squid-sdk/feat-standalone-fallback-source-builders_2026-07-15-10-44.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@subsquid/squid-sdk",
+      "comment": "evm/fallback: EvmPortalDataSourceBuilder and EvmRpcDataSourceBuilder are now full standalone data-source builders (fluent setFields/addLog/… + build()), sharing the query surface with EvmFallbackDataSourceBuilder. Used in a fallback they take connection config only; the fallback injects its shared query, and a source that also sets its own query throws at build() time.",
+      "type": "minor"
+    }
+  ],
+  "packageName": "@subsquid/squid-sdk"
+}
--- a/meta/squid-sdk/README.md
+++ b/meta/squid-sdk/README.md
@@ -64,6 +64,25 @@ const source = new EvmFallbackDataSourceBuilder()
     .build()
 ```
 
+### Standalone sources
+
+`EvmPortalDataSourceBuilder` and `EvmRpcDataSourceBuilder` are full data-source builders in their own
+right — the same fluent `setFields`/`addLog`/… surface as `DataSourceBuilder`. Call `.build()` on one
+directly for a single Portal or RPC source, no fallback involved:
+
+```ts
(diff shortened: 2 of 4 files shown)


abernatskiy:
**(With Claude):** Superseded by #537. Rather than standalone builder classes, downstream sources are now plain tagged config objects (`{type: 'portal'|'rpc'|'source', ...}`) on `EvmFallbackDataSourceBuilder` — no wrapper classes to learn, and the fallback drives a fresh `DataSourceBuilder` per portal source (never wrapping it). Standalone single sources use `DataSourceBuilder` (evm) / `evmRpcStream` (evm/rpc) directly.

PR #543, opened 2026-07-22
fix(evm-rpc): tolerate Optimism deposit transactions (0x7e) that omit…

DamianFigiel:
… nonce

Optimism deposit transactions (type 0x7e) have no signed nonce and some providers (eRPC) omit the field entirely, causing the evm-rpc ingester to reject the whole block with a DataValidationError.

Make Transaction.nonce optional in the evm-rpc validation schema and guard every non-deposit consumer with assertNotNull (the 0x7e encoding path never reads nonce). Propagate the same optionality through evm-normalization, which reads the raw dumps directly, so a missing nonce stays absent rather than being fabricated. The processor archive reader already tolerates it via withSentinel.


tmcgroul:
tried to check if the field is really omitted using dwellir endpoint and as a result couldn't confirm absence of the field. let's consider it a temporal glitch and keep the field required

PR #544, opened 2026-07-22
fix(util-internal): don't wedge ingest head-polling on a hung finalized/head fetch

elina-chertova:
## Cause (proven)

Both `zora-mainnet` EVM hotblocks pods (dwellir **and** uniblock — two independent providers) went silent at the **identical block 49023729 at 17:28:16Z**, stopped emitting head metrics (firing `Portal_Hotblocks_Head_Metrics_Absent_zora_mainnet`), and did **not** recover even though the chain kept producing blocks (verified block 49024000 exists; providers advanced past the stuck point). No crash/restart — the process was **hung**.

Root cause is in `Throttler` (`util/util-internal/src/throttler.ts`), the shared primitive behind every ingest head-/finalized-head tracker:

```ts
async get(): Promise<T> {
    let now = Date.now()
    if (now - this.lastAccess < this.interval) return this.lastValue
    return this.performCall(0)   // <-- returns the in-flight (hung) promise
}
```

`lastAccess` is updated **only on success** (`execute()`). When the underlying fetch hangs or times out for longer than `interval`, the next `get()` falls through and returns the still-pending refresh promise. In `evm-rpc/src/data-source/ingest.ts` the hot loop does `await finalizedHeadTracker.get()` **before** the head poll (`poll.next()`), so a stalled `eth_getBlockByNumber("finalized")` (the pod logged `HttpTimeoutError: request timed out after 30000 ms` on exactly that call) blocks the whole loop. Because the loop runs in a worker, the main-thread `DataService` `for await` then blocks forever with no further logging — matching the observed permanent silence. `enablePrefetch()` is used by every EVM/Solana/Tron/Bitcoin head/finalized tracker, so this can wedge any of them.

## Fix

Honor the prefetch contract: once a value exists, `get()` returns the last known value immediately and lets the (slow/hung) refresh settle in the background, instead of blocking on it. Head polling proceeds; a stale-but-conservative finalized head is safe (only affects the finalized upper bound, which self-corrects on recovery). The first read still awaits the initial fetch, and non-prefetch throttlers are unchanged.

## Tests (red → green)

`util/util-internal/src/throttler.test.ts` (vitest, newly wired into the package's `test` script):
- **the wedge**: prefetch on, first fetch succeeds, subsequent refresh hangs → `get()` must return the cached value promptly. FAILS pre-fix (returns the hung promise → `'HUNG'`), PASSES post-fix.
- first `get()` blocks until the initial value resolves.
- non-prefetch throttler awaits a fresh value when stale (unchanged behavior).

Verified locally: `rush build --to @subsquid/util-internal` ✅, `rush test --only @subsquid/util-internal` ✅. Confirmed red→green by reverting `get()` to the old body (the wedge test fails with `expected 'HUNG' to be 1`).

## Scope / relation to other PRs

This is the *transient-hang → permanent-wedge* fatality (the process should survive a slow/unresponsive finalized endpoint and keep polling head). Distinct root cause from #513 (retry a transient `invalid block height` **error** on the finalized probe) and #523 (portal-client retries transient **null** heads). Complementary, not a duplicate.

## Falsification

If a hotblocks/dumper pod still goes permanently silent (no logs, no restart) after its finalized/head RPC endpoint hangs — i.e. the head loop stops advancing while the chain keeps producing — this fix is insufficient and the block is elsewhere on the head path.

### DIFF
--- a/common/changes/@subsquid/util-internal/alert-fix-UJZrJ3-throttler-head-wedge_2026-07-22-18-11.json
+++ b/common/changes/@subsquid/util-internal/alert-fix-UJZrJ3-throttler-head-wedge_2026-07-22-18-11.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@subsquid/util-internal",
+      "comment": "Throttler: with prefetch enabled, serve the cached value instead of blocking on an in-flight refresh, so a hung head/finalized fetch can no longer wedge ingest head-polling loops",
+      "type": "patch"
+    }
+  ],
+  "packageName": "@subsquid/util-internal"
+}
--- a/common/config/rush/pnpm-lock.yaml
+++ b/common/config/rush/pnpm-lock.yaml
@@ -339,7 +339,7 @@ importers:
         version: file:projects/types-test.tgz
       '@rush-temp/util-internal':
         specifier: file:./projects/util-internal.tgz
-        version: file:projects/util-internal.tgz
+        version: file:projects/util-internal.tgz(esbuild@0.27.7)(tsx@4.21.0)
       '@rush-temp/util-internal-archive-client':
         specifier: file:./projects/util-internal-archive-client.tgz
         version: file:projects/util-internal-archive-client.tgz
@@ -1406,7 +1406,7 @@ packages:
     version: 0.0.0
 
   '@rush-temp/erc20-transfers@file:projects/erc20-transfers.tgz':
-    resolution: {integrity: sha512-a0NVshKVkYr7K4Y1qIl3JBdlGJBmYAk3Cu6Crp4QY9G9vVt0x/E7wqUROArvAo//6PIHge9puyrS/iD5TDLJKA==, tarball: file:projects/erc20-transfers.tgz}
+    resolution: {integrity: sha512-5agkTAoEOVDkvQ53tJQNR6/cp2Wc8jb+OrKQYtVicfiq2Qk61vsliFkcZxEOOlvgAhI9TG1wgQkC2hXM5OBQPQ==, tarball: file:projects/erc20-transfers.tgz}
     version: 0.0.0
 
   '@rush-temp/evm-abi@file:projects/evm-abi.tgz':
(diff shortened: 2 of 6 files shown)


eldargab:
The contract of the `Throttler` is -  Allow the value to be stale for no longer than `.interval`, 
and it is used in many places.

This PR changes the contract and I can't calculate the consequences out of my head.

Depending on how relevant the AI analysis actually is, the fix should be related to either 
the Ingest loop or "Worker thread - Main thread" error propagation.

SQD-Trevor-Agent:
Second independent production instance of this exact wedge — a different network, a different provider, and a different trigger than the zora timeout case, but the same `Throttler.get()` hang.

**Instance: `ink-mainnet` alchemy EVM hotblocks**
- Pod `evm-alchemy-ink-mainnet-hotblocks-service-7f5cf7cdfb-zcgjg` started `2026-09-16T07:52:12Z`, emitted ~35 log lines, and went **completely silent after `2026-09-16T07:53:01Z`** — no crash, no restart, HTTP server (`:3000`) still up and being scraped.
- Its `sqd_hotblocks_last_block` is **frozen at block `56045406`** while the chain advanced to `~56206110` (the sibling `uniblock` replica is healthy at the head, so the chain and the dataset are fine). `sqd_hotblocks_last_block_lag_ms` is also frozen at the value it held the moment it hung (`1351560`), which is why the lag alert reads a stale ~22 min even though the pod is now ~44 h behind.
- The final log lines are all `eth_getBlockByNumber(["finalized", false])` returning HTTP **429 “exceeded compute units per second capacity”**, with `will pause new requests for …ms` escalating `10 → 100 → 500 → 2000 → 10000 → 20000`, then silence.

**Why it matches this PR's root cause**
The 429 path in `util/rpc-client/src/client.ts` keeps the finalized-head request pending across the connection-error backoff (re-enqueue + escalating `backoff()` / `waitForConnection()` loop) rather than settling it. So the `Throttler`'s `fn()` never resolves, `execute()` never updates `lastAccess`, and every subsequent `get()` falls through to and returns the still-pending promise. `evm/evm-rpc/src/data-source/ingest.ts:73` (`await finalizedHeadTracker.get()`) then blocks the `jobs()` generator forever — exactly the permanent silent hang described here. Trigger differs (a sustained 429 CUPS storm on the Alchemy key shared across ~17 archive networks vs. a 30 s HTTP timeout), mechanism is identical.

**On @eldargab's contract concern**
Agreed that returning an arbitrarily-stale value from `get()` weakens the “stale for no longer than `interval`” contract that other call sites rely on. An alternative that preserves that contract while still removing the permanent-wedge fatality: **bound the refresh instead of returning stale**. e.g. an opt-in per-fetch timeout on the throttler (default off, so every existing caller is byte-for-byte unchanged) — when the underlying `fn()` exceeds the bound it **rejects** and clears `pending`, so `get()` rejects, the head loop errors and retries (or the process exits and k8s restarts) instead of hanging silently forever. The value is then never older than the bound (contract kept), and a hung finalized endpoint can no longer wedge ingest. Happy to spin the throttler change that way if you prefer it to the stale-read approach.

Note the branch is currently `CONFLICTING` and needs a rebase onto master.

PR #561, opened 2026-08-31
perf(fallback): poll liveness with eth_blockNumber instead of a block lookup

abernatskiy:
## Summary

The fallback source's freshness machinery (lag reference, liveness, stall recovery) polls standby heads once per batch at tip pace — the batch cadence outruns the 5s head cache — and each poll fetched a full block reference via `eth_getBlockByNumber`, of which only the **number** was ever consumed. Users watching provider dashboards see a steady stream of block lookups from an idle standby.

- `DataSource` gains an **optional** `getHeight(): Promise<number>` — a number-only head poll for liveness/lag consumers. Sources that don't implement it are unaffected (the fallback falls back to `getHead().number`).
- `EvmRpcDataSource` answers it with `eth_blockNumber` — typically the cheapest (often unmetered) call a provider offers.
- The fallback's `headWithTimeout` prefers `getHeight` when the source has one.

Solana is untouched: its `getHead` already rides `getLatestBlockhash`, which is a cheap native call.

## Test plan
- New: fallback prefers `getHeight` over `getHead` for the poll (asserts zero `getHead` calls on a source offering both); `EvmRpcDataSource.getHeight` delegates to `eth_blockNumber` only.
- Full suites green: fallback (75), evm-rpc (213), squid-sdk meta (175).

The sibling change in the Pipes SDK additionally gates its boundary poll; this repo's fallback already gates every poll path (`laggingTooFar` early-returns with lag off, `probeHigherPreference` is a no-op at index 0, the stall poll fires only past the staleness window), so only the call cost needed fixing here.

https://claude.ai/code/session_01W2HQnYVvTkKkVPJCB6pjn6

### DIFF
--- a/common/changes/@subsquid/evm-rpc/perf-cheap-head-polls_2026-09-01.json
+++ b/common/changes/@subsquid/evm-rpc/perf-cheap-head-polls_2026-09-01.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@subsquid/evm-rpc",
+      "comment": "EvmRpcDataSource implements the new optional getHeight() with eth_blockNumber, so freshness head polls no longer cost a full eth_getBlockByNumber lookup",
+      "type": "minor"
+    }
+  ],
+  "packageName": "@subsquid/evm-rpc"
+}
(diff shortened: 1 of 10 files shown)


abernatskiy:
**(With Claude):** Follow-up commit: a live smoke test (method-counting proxy in front of the RPC standby, tip-following run) caught that `EvmRpcStreamDataSource` — the adapter the fallback actually holds — did not forward the new optional `getHeight`, so polls silently stayed on `eth_getBlockByNumber`. Now forwarded, with an adapter-level regression test. Measured after the fix: the recurring standby traffic is `eth_blockNumber`. The identical omission existed in the Pipes SDK lazy RPC wrapper and is fixed there too.

PR #568, opened 2026-09-14
fix(substrate-processor): build fails on extrinsic extension fields

mo4islona:
`master` does not build since e5a75484: `@subsquid/substrate-processor` fails with TS2739 in `src/selection.ts`, which breaks unit tests and the release workflow.

e5a75484 added `extensionVersion` and `extensions` to the `substrate-data` extrinsic, for tip parsing of v5 general extrinsics. `FieldSelection['extrinsic']` is derived from the keys of that type, so the selection validator no longer matches it.

Neither `ds-rpc` nor `ds-archive` copies these fields to processor extrinsics, so adding them to the validator would expose selectable fields that are always empty. This change excludes them from selection instead; exposing them is left to a follow-up that also maps them.

Checked: `tsc` fails on master with the CI error and passes with this change; `rush build --from @subsquid/substrate-processor` succeeds.

https://claude.ai/code/session_01CCCpbRHXFjW3PT2wPYcHWq


mo4islona:
Superseded by 2576730a on master, which excludes the same fields from field selection.

